PR #42, opened 2015-01-28
MyDriver structure renamed to MyRFDriver

fallberg:
The MyDriver scheme is radio specific, so it is renamed
accordingly to permit other driver schemes to coexist with it
without confusion.


fallberg:
As this is unrelated to the signing development I am creating a separate pull request for this.
My ambition is to make a similar construct for signing back-ends and this change will make it a bit more clear for other developers how this structure is intended to be used. So this is more "in preparation" for the signing commits which will eventually come.

PR #43, opened 2015-01-29
fix a case where parent distance from repeater get stuck at 255

mapnull:
I had a repeater stuck at 255 in nc.distance because the gateway went down for a short time and miss the I_FIND_PARENT message when got back.


henrikekblad:
Great job finding this Marcello!

Would you mind adding your name to the CLA so I can merge it?

mapnull:
I could only add to mysensors/Vera, the mysensors/Arduino is giving this error message from CLAHub:
"We're sorry, but something went wrong."

henrikekblad:
Ok, might be a bug in clahub! Bugs everywhere ;)
Should be ok if you added your name to the other one.

Thanks again for the commit.

PR #44, opened 2015-01-30
fix a possible parent repeater loop

mapnull:
If a repeater responds to his parent node, this can cause a loop between the two.

(I managed to sign for both CLAs now)


henrikekblad:
Thanks!

PR #46, opened 2015-02-03
Bugfix & Clear routing table extendend

tekka007:
I_CHILDREN + C: clears routing table, parentnode and distance, followed by findParent(). This command is useful to re-discover network topology in case of repositioned or offline nodes.


henrikekblad:
Thanks for the PR.

Did you also experience problems signing the CLA (like a few others lately)?

henrikekblad:
Ok, added an issue over at CLAHub.. In the meantime If are unable to sign you could perhaps reply to this pull request and say "I Agree to CLA". 

Sorry for the inconvenience. This is all pretty annoying.

tekka007:
Yup, same problem here. Nevertheless, i agree to the CLA :)

PR #47, opened 2015-02-03
Fetch id through repeaters

mapnull:
perhaps the documentation I wrote should be improved?


henrikekblad:
The doc is ok. Thanks for the PR.

PR #48, opened 2015-02-03
get custom string from msg that isn't a stream type

mapnull:
I was thinking in the possibility of sending a C_SET message with a P_CUSTOM payload.


henrikekblad:
Yep, could be useful when serializing custom data from sensors out to controller. E.g. IR-codes.

PR #50, opened 2015-02-04
led inversion hack

rhooper:
This little hack makes it so you can invert the gateway LEDs if you wired them up such that they're always on unless there's activity.

It may be better refactored to use a define to save RAM.


henrikekblad:
Thanks rhooper,

I might wait on pulling this in. Really low prio and only needed if leds are connected wrong without pullups (if I understand it correctly).

PR #51, opened 2015-02-05
MysensorMicro example, and extra libraries

tbowmo:
Hi Henrik

Here is my addition to the libraries etc. used by the mysensor Micro (at the moment..) might change a little when I start to look at the next revision of the board :)


bblacey:
Thomas, the SI702 sample library you used carries with it a GNU GPL license.  We tend to favor less stringent acceptable use licenses such as BSD, MIT, etc.  How about using this MIT-licensed library instead - https://github.com/electricimp/reference/tree/master/hardware/Si7021.  

That said, awesome work overall on the MySensors Micro ;)

tbowmo:
Hi Bruce

It's not an arduino library, that you linked to? So won't fit our purpose

Anyway I'll look around and see if I can find a better licensed one.

/Thomas
Den 06/02/2015 17.14 skrev "Bruce Lacey" <notifications@github.com>:

henrikekblad:
I'll merge this for now (did a quick check and we're already having quite a few GPL libraries bundled). 
We can maybe make a later attempt to clean this up for the whole project.

PR #52, opened 2015-02-06
MYSBootloader 1.0

tekka007:
Bootloader for (direct) OTA FW updates. Occupies 1834 bytes
.hex is compiled for 16Mhz. Bootloader was tested in this frequency
range; 128kHz (intRC) to 16Mhz(XTAL) at 3.3 and 5V.
For OTA bootloading, MYSController was used, NodeJS Controller not
tested.


henrikekblad:
Great job on this one tekka007!

PR #53, opened 2015-02-06
Request/send node info only after receiving the id

mapnull:
Maybe line 81 should be placed some elsewhere?
Why we need to open a pipe in WRITE_PIPE ?


henrikekblad:
Thanks, 

Maybe it would be good to break out the presentation/config request to a method? Could you at least check if it affects code size?

mapnull:
I will send another PR using method

PR #57, opened 2015-02-07
Fixing endless loop

tekka007:


henrikekblad:
Embarrassing x 2.

Thanks!

PR #58, opened 2015-02-08
simplify wait(), substitute waitForReply(), adding watchdog resets

tekka007:
Simplify wait(), added watchdog reset to wait() and waitForReply().
Substituted all waitForReply() for wait(2000) to reduce code
size/complexity


henrikekblad:
waitForReply could probably be removed as nothing actually uses it now.

tekka007:
Yes, true.

PR #62, opened 2015-02-11
Accept a new parent only when autoFindParent is 1

mapnull:
In a gateway, I_FIND_PARENT_RESPONSE messages was returning as true by
process(). With this commit it will return as false, is this the right
expected behavior?


henrikekblad:
Yes, seems more correct.

mapnull:
I initially thought fixing the case of a common node could rewrite his father id even though set to be static, later I saw that could be extended to the gateway with the autoFindParerent variable.

PR #64, opened 2015-02-13
process() uses sendRoute() for routing decisions

mapnull:
To keep routing decisions in one place, and avoid rewriting the same code to add failedTransmissions count to process()


mapnull:
let me redo, forgot to add some code

PR #83, opened 2015-02-27
fix multiple mqtt clients with openhab

mapnull:
There is probably a better solution, this was the simplest I came up with.
Changed some variables types from uint8_t to byte to make simple for common arduino user.


henrikekblad:
http://blog.oscarliang.net/arduino-difference-byte-uint8-t-unsigned-cha/

PR #86, opened 2015-03-13
Split out PinChangeInt.h into header and cpp source

peterdey:
I'm using MySensors on a DevDuino node: http://www.seeedstudio.com/wiki/DevDuino_Sensor_Node_V2.0_%28ATmega_328%29

These nodes have a hardware button connected to D4, which alas, has no hardware interrupt available.  You can hook an interrupt up to D4 using PinChangeInt, to wake the node from sleep.  Useful for rebooting the node without pulling the batteries, or for waking it to force it to transmit a new set of readings.

You can't #include your own copy of PinChangeInt, because MySensors has all the code contained within a header.  You'll get lots of "already defined" errors when linking.  Moving the actual code into its own CPP file allows you to legally #include PinChangeInt.h twice, without hitting errors.


henrikekblad:
Hi @peterdey 

Thanks. If you look in the development-branch the PinChangeInt is already removed from the core library.

peterdey:
Whoops, sorry!!

henrikekblad:
No problem! Thanks for the effort anyway.

PR #87, opened 2015-03-13
Clearing EEPROM values to start with blank state

binodmainali1:


henrikekblad:
Discussion continues here:
http://forum.mysensors.org/topic/1102/

PR #90, opened 2015-03-18
Bugfixes and whitelisting support in signing backends

fallberg:
Highlights of the changes in this pull request:
* Whitening of nonces (strengthens the quality of the random values)
* ATSHA personalizer updated to further simplify copy+paste
* Backend memory footprint reduced
* BUGFIX: EEPROM storage of signing request table now corrected
* Signing can be turned of by removing MY_SIGNING_FEATURE definition in MyConfig.h
* PinChangeInt library patched to only support one pin (and not use new). Reduces program size significantly.


henrikekblad:
Great

PR #92, opened 2015-03-23
Modified WaterMeterPulseSensor Example

gizmocuz:
Modified watermeter pulse example
- Current master code would send unlimited gateway get VAR_1 requests at the fasted interval possible
- New code start counting pulses, when received the VAR_1, it will be added. No missing pulses
- Cleaned up the code, corrected some spelling/comments


henrikekblad:
Great!

PR #100, opened 2015-04-10
Initial Support for RFM12B Modules

xpix:
Hello, this changes are untestet ... but for the first time :) Please check this for support the RFM12B Modules, i use the LowPowerlab RFM12B Modules for this change. 

Hope this helps :)
Frank


henrikekblad:
Great addition Frank, thanks!

Do you have the possibility to make some sanity testing (as I don't have any RFM12B modules) before I take this in? 

Also you might wanna rollback to the defaults in SerialGateway.ino for the PR.

What is the spec differences between FF12B vs RF69?

garyStofer:
@xpix : I was just thinking of doing this for the RFM22B module.  Do you know what the differences are between the RFM12B and the RFM22B 
@henrikekblad Biggest diff between RFM69 and 12/22 is that 69 has encryption built in and has the RX/TX antenna switch built in (afaik)

henrikekblad:
Thanks for the update @xpix. Did you get a chance to test it on your RFM12Bs?

xpix:
Sure, i want to test this week and at the weekend. I hope i'll bring this to fly. If u want i can close this PR and send a new one when i'm finished?

xpix:
So, yesterday i run in an freeze problem with Arduino Unos and this Modules. Only my small Arduino Pro Minis want to work together with RFM12B Modules. I have a lot of experiences with RFM12B but i worked only with the jeelib. Now i want to use the RFM12 lib from lowpowerlabs, cuz know you use the RFM69 from the same author :) I guess at the weekend i'll bring this project to fly :)

henrikekblad:
Great, thanks for the update.

xpix:
Ok, i worked the hole weekend on this modules and now i give up :) Ok here my Problem: 

> Node request a new Node id and Gateway don't answer cuz "   }
> else if (sender == GATEWAY_ADDRESS) {" doesn't match

**Sensor**:
req id
send: 255-255-0-0 s=255,c=3,t=3,pt=0,l=0,sg=0,st=ok:
**SerialGateway**
0;0;3;0;9;gateway started, id=0, parent=0, distance=0
0;0;3;0;14;Gateway startup complete.
0;0;3;0;9;read: 255-255-0 s=255,c=3,t=3,pt=0,l=0,sg=0:
0;0;3;0;9;Sender 255, Gateway: 0, 
255;255;3;0;3;

Problem, Row 552 in MySensor.cpp
			} else if (sender == GATEWAY_ADDRESS) {

May somebody can help me, tommorow i'll ask the forum.

henrikekblad:
Hi, gateway won't answer any id requests. You normally have a controller behind it which answers these.  

You could just do some tests with static id for the node (if you don't want to play with the controller parts).

On your node use the following to set id 42:
```
gw.begin(NULL, 42); 
```

xpix:
Hmm, the communication between nodes fly. The node get the parent-id, thats success. Ok, first part run :)

xpix:
Yesterday, i was too tired. Ok, today i got new energy and i'll try your tip. Now i use the JeeLib library, cuz i have some experiences and the lib are very stable. The bad point, only support 1-30 Nodes but 255 Networks. The good point, you can mix your network with RFM12 and RFM69 nodes. They can communicate with each other.

henrikekblad:
What is the reason it only support 30 nodes? Are they using frequency as "id"? This will be problematic for the controller which do _think_ they can hand out ids between 1-254.

xpix:
The Header in jeelib use only a 5bit number to address the nodes, but we
can use (later) one byte from payload to adress to 255 Nodes. May the RFM12
from lowpowerlab are better, but this lib can only address 1-127 Nodes:

https://github.com/LowPowerLab/RFM12B

If it possible to say the controller "Only 1-30 Nodes"? Other question, can
u give me an example in code where the place to set a nodeid if requested?

*--*
*Frank Herrmann    | Software Entwicklung*
+49 172 600 40 61 | xpixer@gmail.com
Neudorfer Weg.10  | 68753 Waghäusel

2015-04-20 9:41 GMT+02:00 Henrik Ekblad <notifications@github.com>:

henrikekblad:
Here is an example of how a controller could respond to id-request:
https://github.com/mysensors/Arduino/blob/master/NodeJsController/NodeJsController.js#L404

Right now there is no option for gateway to inform controller of valid id-range. That would require some new internal configuration command between them.

xpix:
Hello Henrik,

now i understand the code. The SerialGateway get a request for new id and
the controller watch the output from gateway and answer this question with
a new ID-Command. Correct? I guess i misunderstand the code and thought the
gateway managed the free nodeid db :)

Ok, for the first step without controller we can set manual the node id in
SensorNode:

  #define NODEID 3
  ...
  gw.begin(NULL, NODEID);

I'll test this today and i guess this will work.

*--*
*Frank Herrmann    | Software Entwicklung*
+49 172 600 40 61 | xpixer@gmail.com
Neudorfer Weg.10  | 68753 Waghäusel

2015-04-20 11:05 GMT+02:00 Henrik Ekblad <notifications@github.com>:

xpix:
Hello henrik, ok the nodes run with the last commited version and a setted NodeID. I add also some examples, if you dont need them then remove this please. 
One Issue: After some seconds the SerialGateway crashed and try to make a reset. I don't know why, i test on Arduino Mini 3.3 V and Jeenode 3.3V, same result. May you know what are happend?
```
 0;0;3;0;9;gateway started, id=0, parent=0, distance=0
 0;0;3;0;14;Gateway startup complete.
 0;0;3;0;9;read: 22-22-0 s=255,c=0,t=17,pt=0,l=6,sg=0:1.5 b1
 22;255;0;0;17;1.5 b1
 0;0;3;0;9;read: 22-22-0 s=3,c=0,t=0,pt=0,l=6,sg=0:1.5 b1
 22;3;0;0;0;1.5 b1
 0;0;3;0;9;ga
```

May you know what are happend?

henrikekblad:
> now i understand the code. The SerialGateway get a request for new id and
> the controller watch the output from gateway and answer this question with
> a new ID-Command.

Yes. 

Regarding crash... Hmm. Not anything I've experienced. I guess you don't see this when using NRF24L01 radios with its driver?
Is the RFM12-radio power hungry? Tried adding a capacitor over vcc/gnd on radio? Last probable cause is probably a buffer over-write somewhere in the new code/driver.

henrikekblad:
Hi @xpix, 

What do you suggest to do with the RFM12 drivers? Guess you never got it stable?

xpix:
Hi Henrik, sadly not ... my knowledge are not enough. As far as i remember after some seconds the gateway crashed. I give up at this time, sorry.

henrikekblad:
No problem. Thanks for trying. Just wish I could help you more resolving it.

PR #101, opened 2015-04-12
Fix for debug messages in MyTransport and removing executable bit from sources

thozza:
Hi.

While rebasing the dynamic discovery changes I found out that the debug messages in MyTransport classes were not printed. The problem was that MyConfig.h in which the DEBUG macro is defined, was not included. Another problem is that debugPrint() is defined in MyHw. I was trying to rework the MyTransport to get MyHw instance as an argument and then use it for logging. This turned out to be too much of a hassle and the easier approach was to implement simpler debugPrint in the MyTransport.cpp. Yet another approach would be to use Serial.println() directly.

I also removed the executable bit from source files, since it does not make sense...


henrikekblad:
Hmm.. the only debug print I could find was this:
https://github.com/mysensors/Arduino/blob/development/libraries/MySensors/MyTransportNRF24.cpp#L36

Maybe better just to let init() return false when it fails. And put the debug print/while(1) in the MySensors-class?

thozza:
Right, the only place where it is used is in the MyTransportNRF24.

The return value is a better Idea. I'll rework the patch and update the PR.

thozza:
OK, the fix is updated... ;)

henrikekblad:
Thanks!

PR #103, opened 2015-04-18
[Enhancement] AutoID for Serial Gateway

funky81:
Following is the patch for Auto ID in Serial Gateway.
Serial Gateway will start from ID 1 till 254 where 255 is the ID for undefined id (temporary) and ID 0 is for the gateway itself

### DIFF
--- a/libraries/MySensors/MyGateway.cpp
+++ b/libraries/MySensors/MyGateway.cpp
@@ -222,6 +222,17 @@ void MyGateway::setInclusionMode(boolean newMode) {
     }
 }
 
+inline MyMessage& build (MyMessage &msg, uint8_t sender, uint8_t destination, uint8_t sensor, uint8_t command, uint8_t type, bool enableAck) {
+	msg.destination = destination;
+	msg.sender = sender;
+	msg.sensor = sensor;
+	msg.type = type;
+	mSetCommand(msg,command);
+	mSetRequestAck(msg,enableAck);
+	mSetAck(msg,false);
+	return msg;
+}
+
 void MyGateway::processRadioMessage() {
 	if (process()) {
 	  // A new message was received from one of the sensors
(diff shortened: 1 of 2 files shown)


thozza:
Out of curiosity, I thought the controller is the one assigning Node IDs? With this change, the gateway would always be assigning IDs to nodes.

henrikekblad:
Hi @funky81,

Thanks for the patch but if gateway would start handing out ids we would also have to create control commands for un-assigning id's from the controller side etc. Also, if a user would switch gateway he'd lose the count and gateway would hand out duplicates to new nodes. 

If we would put more "intelligence" in the gateway I would like to se a more complete set of controller<->gateway commands and a gateway that keeps track of sensor net (keep a list of known sensors/actuators). 

Unfortunately the gateway storage on 328p is already over the limit (with message signing enabled) and we would have to step up hardware wise on the gateway to cope with this kind of things.  

The MQTT gateway is today handing out ids but this is not good and will probably change in the future. 

Are you using the gateway with any controller in your setup?

funky81:
I've just know that ID should be given by Gateway. If it work like that,
how is the mechanism?
Currently I'm using Pimatic and as far I check in the source code, it still
doesn't support Auto ID for MySensor.

Thanks

funky81:
I've just realize this, any gateway that support this mechanism? Coz my
gateway, Pimatic, still doesn't support this

Thanks

funky81:
Sorry guys, any response for this? 
If not, then let me have it on my own branch.

Thabks

henrikekblad:
Sorry for a late response.
You'll probably have to have this in your branch until Pimatic supports id-handout.

funky81:
Thanks for your clarification Hek.
May I know, which gateway that already support auto ID? Need to implement
this.

Thanks

henrikekblad:
Most of them. Didn't know that pimatic hadn't implemented it. Maybe you could help them to get it in place?

funky81:
Ok, Thanks Hek. I'll close this PR.

Thanks

PR #104, opened 2015-04-20
MyTransportNRF24::available() could return true if initialization failed

thozza:
This caused the sketch to process message that didn't arrive.

Signed-off-by: Tomas Hozza <thozza@gmail.com>

### DIFF
--- a/libraries/MySensors/MySensor.cpp
+++ b/libraries/MySensors/MySensor.cpp
@@ -102,7 +102,7 @@ void MySensor::begin(void (*_msgCallback)(const MyMessage &), uint8_t _nodeId, b
 
 	// Setup radio
 	if (!radio.init()) {
-		debug(PSTR("radio initialization failed!"));
+		debug(PSTR("radio initialization failed!\n"));
 	}
 
 #ifdef MY_SIGNING_FEATURE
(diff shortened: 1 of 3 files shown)


libraries/MySensors/MySensor.cpp
```diff
@@ -102,7 +102,7 @@ void MySensor::begin(void (*_msgCallback)(const MyMessage &), uint8_t _nodeId, b
 
 	// Setup radio
 	if (!radio.init()) {
-		debug(PSTR("radio initialization failed!"));
+		debug(PSTR("radio initialization failed!\n"));
```
henrikekblad: No sure it is worth continuing here if radio initialisation fails. 

How about just do a while(1); ?

thozza: Well, that is basically what happens then in the process() function. But yes, this approach would be also functional. I can rework it that way if you want... ;)

henrikekblad: I could do it if you want. ;)

But the fundamental question is what we want here? Is it of any value to continue? In some use cases it might be but I cannot come up with one now;). Hmm... In this case the radio is probably mis-wired and we should probably just stop the execution to alert the user as it is none-recoverable.

thozza: Right. I was thinking that it may be reasonable to try the radio initialization again if previously failed (in the loop()). E.g. I have board Devduino v2.0 on which I have to unplug the RF24 module when programming the Atmega. So once the board is reset after the programming I may plug in the RF24 late, so another initialization may succeed.

Other than this situation, the failed initialization is commonly due to bad wiring so there is rally not much sense in executing any other code.

So it depends if these extra lines of code and extra program memory used by it is worth it.

henrikekblad:
Ok, committed the while-thingie. Thanks.

PR #106, opened 2015-05-04
fix check for NULL buffer in getString()

mapnull:


henrikekblad:
Thanks, yep, might be good to do a null check here.

PR #109, opened 2015-05-06
Add LEDs blinking into base MySensor class without any library dependency

thozza:
Hi.

As we discussed in PR #66 I added the possibility to blink LEDs into the base MySensor class. I tested it with simple temperature sensor based on Arduino Mini Pro and with Serial Gateway. I also modified the Ethernet Gateway. I left the MQTT gateway as is, since I don't have testing setup for MQTT and the Gateway is not using the common GatewayUtil.h as the other two gateways.

Next step... move the inclusion mode feature into the base class (most probably reintroduced MyGateway class) without the interrupts library. And then finally rebase the dynamic discovery ;)

### DIFF
--- a/libraries/MySensors/MyConfig.h
+++ b/libraries/MySensors/MyConfig.h
@@ -46,6 +46,26 @@
 
 
 /**********************************
+*  Information LEDs blinking
+***********************************/
+// This feature enables LEDs blinking on message receive, transmit
+// or if some error occured. This was commonly used only in gateways,
+// but now can be used in any sensor node. Also the LEDs can now be
+// disabled in the gateway.
+
+// #define WITH_LEDS_BLINKING
+
+// default LEDs blinking period in milliseconds
+#define DEFAULT_LED_BLINK_PERIOD 300
+// The RX LED default pin
+#define DEFAULT_RX_LED_PIN 8
+// The TX LED default pin
+#define DEFAULT_TX_LED_PIN 9
+// The Error LED default pin
+#define DEFAULT_ERR_LED_PIN 7
+
+
+/**********************************
 *  Message Signing Settings
 ***********************************/
 // Disable to completly disable signing functionality in library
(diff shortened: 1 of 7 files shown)


andriej:
You don't even know how happy I am with this PR :-)
Signed, LED lover.

henrikekblad:
@thozza, this looks great!  So sweet to throw away the MSTimer dependency. 

Yes, we have to reintroduce the base class. Maybe we could have a chat about that on skype/hangout?. Really need to clean up and make the gateway-class more like the MySensors class where we can configure both Transport (Ethernet, Serial, ESPWifi) and protocol (MySensors, MQTT, Json?). 

We @tekka007 has already broken out the Serial-parser stuff but this could probably be even more generalised into a "Protocol"-implementation that can convert between MyMessage and some string depending on protocol implementation.

(sorry for the long comment)

thozza:
@henrikekblad sure, we can discuss it. I added you in my circles on G+. I'm all for making the code more clean and modular ;)

thozza:
@bblacey the reason for not using any library is that Henrik asked for it explicitly in PR #66, since it was causing issues with some sketches. I think that not using the library also decreases the size of compiled firmware, since MySensors needs just specific functionality. Also LEDs can now be used with regular sensor nodes, not only Gateways. Also if not needed in Gateway, the functionality can be completely left out and not compiled in.

As for the default PINs, I simply used what was already used in the SerialGateway and also EthernetGateway (e.g. https://github.com/mysensors/Arduino/pull/109/files#diff-8507ab9ef1b10c2416fdbfbdacff8766). I don't have any problem with changing the defaults, personally I always define PINs explicitly in my projects. Maybe we could use some of the Analog PINs... I don't have any strong opinion about this.

bblacey:
Thomas, it seems that it would be easier for people who have already built gateways with LEDs to maintain the default 6,5,4 pins for RX, TX, and ERR respectively.  I thought about the analog pins too but they are useful for analog sensors like PIR, etc. so it might be good to leave those open for sensors.  Anyway, very cool that you implemented this and great job doing it!  :)

PR #111, opened 2015-05-12
Cleaned up Sensebender sketch, and implemented running average

tbowmo:
I have implemented Running Average on temperature and humidity, in order to avoid "flickering" in the measurement values to trigger a transmission, as I specially see the humidity changing back and forth between two values, f.eks. 40 and 41% it forced a transmission before.

Now it calculates the average of the last two measurements, and compare that to the new measurement taken, and determines if it should transmit from the result.

(@bblacey could you please verify on your samples, if you've got the time :)).

### DIFF
--- a/libraries/MySensors/examples/SensebenderMicro/SensebenderMicro.ino
+++ b/libraries/MySensors/examples/SensebenderMicro/SensebenderMicro.ino
@@ -29,7 +29,7 @@
 #define CHILD_ID_HUM   2
 
 // How many milli seconds between each measurement
-#define MEASURE_INTERVAL 1000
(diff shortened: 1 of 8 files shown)


bblacey:
Thomas, I noticed that the clock_prescale_set() in setup() is commented out - is that what you intended?

+//  clock_prescale_set(clock_div_8); // Switch to 1Mhz right after powerup.

tbowmo:
Aarghh.. found a couple of other mishaps in the code (Henrik pointed me at one of them :))

I'll close this one, and make a new PR with code corrected..

PR #113, opened 2015-05-23
Add Ethernet Transport driver for Gateway

thozza:
Current state:
* gateway can work as client - must define the controller IP and port
* supports UDP or TCP (the classis EthernetServer)
* static IP or DHCP

The TCP + DHCP can not fit on 328p, due to Arduino IDE compiling and linking a lot of garbage that is not really needed.

I'll look at the **server** mode and dynamic discovery later.


henrikekblad:
Great! The gateway sketch is sooo clean ;)

PR #122, opened 2015-06-07
Added battery reporting as a sensor value

tbowmo:
Added a define (and code), to enable battery voltage measurement reporting, as a normal sensor value. This will enable people to log battery voltage over time.

Default is disabled for ths functionality.

### DIFF
--- a/libraries/MySensors/examples/SensebenderMicro/SensebenderMicro.ino
+++ b/libraries/MySensors/examples/SensebenderMicro/SensebenderMicro.ino
@@ -3,7 +3,7 @@
 //
 // If A0 is held low while powering on, it will enter testmode, which verifies all on-board peripherals
 // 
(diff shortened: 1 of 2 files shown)


libraries/MySensors/examples/SensebenderMicro/SensebenderMicro.ino
```diff
@@ -20,6 +20,9 @@
 // Define a static node address, remove if you want auto address assignment
 //#define NODE_ADDRESS   3
 
+// Uncomment the line below, to transmit battery voltage as a normal sensor value
+//#define BATT_SENSOR    199
```
thozza: Why didn't you add this to the MyMessage.h?

tbowmo: This is specific for the sensebender micro...
Den 07/06/2015 22.36 skrev "Tomas Hozza" <notifications@github.com>:

PR #123, opened 2015-06-17
Some changes for the DallasTemperatureSensors.ino and BatteryPoweredSensor.ino

Kolbi:


Kolbi:
Made debugging information for BatteryPoweredSensor optionally throught settings.

tbowmo:
Just a thought, why not use #ifdef DEFINE, instead of #if DEFINE = 1 ?

Kolbi:
Just thought it's easier to understand for newbies to set the value to 1 or 0 instead of uncommenting or commenting the define.

PR #129, opened 2015-06-28
Non-blocking Dallas sensors temperature conversion

roboprint:
Updated DallasTemperature and OneWire libraries for non-blocking `requestTemperatures()` call. This allows deep sleep or process incoming messages during temperature conversion (up to 750 ms).

### DIFF
--- a/libraries/DallasTemperature/DallasTemperature.cpp
+++ b/libraries/DallasTemperature/DallasTemperature.cpp
@@ -3,369 +3,491 @@
 // License as published by the Free Software Foundation; either
 // version 2.1 of the License, or (at your option) any later version.
 
+// Version 3.7.2 modified on Dec 6, 2011 to support Arduino 1.0
+// See Includes...
+// Modified by Jordan Hochenbaum
+
 #include "DallasTemperature.h"
 
+#if ARDUINO >= 100
+#include "Arduino.h"
+#else
 extern "C" {
-  #include "Arduino.h"
+#include "WConstants.h"
 }
+#endif
 
 DallasTemperature::DallasTemperature(OneWire* _oneWire)
-  #if REQUIRESALARMS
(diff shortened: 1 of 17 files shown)


jendrush:
Hello @roboprint , i have strange problem with DS18B20 and MySensors. Recently i've changed version of the MySensors library, and i know that MySensors updated DS18B20 library aprox mont ago. I have problem with my sensor in freezer, there are temperatures around -26 degrees C. Software tells me there is -2, -3 (aprox) degree C, and i know from another sensor there is -25,-26(aprox) degree C. Maybe my latest observation tells you something. When i heat sensor in my hands for few seconds (5-10) it shows proper temperature, look at this screenshoot http://screenshooter.net/1288732/jjxfdbb . There are three tests, left, and right are heating for 5-10 seconds, and middle is heating longer, maybe a minute. I think something could be wrong with this newest update becouse earlier i was having perfectly good readings within maybe 1,5 year.

roboprint:
@jendrush can you downgrade mysensors library (or onewire + dallas only) and test ?

jendrush:
I am planning to do this tomorrow. I will let you know about result.

jendrush:
Uploaded older sketch to node, and same thing is happening. Don't know what is going on..

roboprint:
Try different dallas sensors and try simplest dallas sensor test sketch (read temperature + output to serial console)

jendrush:
Ok, i've made some tests. Two different sensors, two different arduinos, and result the same in Domoticz. Simple sketch is showing proper temperature http://screenshooter.net/102499302/saitpph .

roboprint:
Can you hardcode -22 temp instead of reading from sensor in mysensors sketch?

jendrush:
Mysensors is very nice library, i will try to do this work. I will test few things tomorrow.

PR #133, opened 2015-07-08
Refactored forecast algorithm

windkh:
- reduced footprint by storing only the last few samples instead of all 180
- fixed wrong calculation: 65/1023 factor replaced with a conversion factor hPa to kPa
- fixed error when 180 minutes were reached: sketch started at 5 min instead of 6 triggering a wrong average calculation

Cleaned up code
- removed unused variable
- fixed unit Pa to hPa in serial print
- introduced constants and enum for better readability

Signed-off-by: windkh <karl-heinz.wind@web.de>

I agree to the CLA!


henrikekblad:
Nice! Thank you!

PR #134, opened 2015-07-11
Changed the extended_fuses from 0xFE to 0x06 (set unused bits 3-7 to …

bblacey:
Reviewed this change with @tbowmo in Hangouts and he agreed with the change...

>Thomas, do you mind if I create a pull request to change the extended_fuses from 0xFE to 0x06?  The upper bits are inaccessible by certain programmers and not writable anyway so this would create broader compatibility.  I always have to change it when I want to burn a bootloader with the ATMEL Dragon or USB Tiny ISP.
In boards.txt

>>Thomas Bowman Mørch
Be my guest Bruce


bblacey:
The second commit cleans up the WITH_LEDS_BLINKING option as discussed in this thread https://github.com/mysensors/Arduino/pull/109

PR #136, opened 2015-07-16
Allow disabling of LED/Inclusion pins

Yveaux:
LED/inclusion pins will always be controlled, even when not configured
to use them.
This fix will ignore pins when a negative pin number is passed.

### DIFF
--- a/libraries/MySensors/MyGateway.cpp
+++ b/libraries/MySensors/MyGateway.cpp
@@ -14,17 +14,20 @@
 #include "utility/PinChangeInt.h"
 
 
-uint8_t pinRx;
-uint8_t pinTx;
-uint8_t pinEr;
+int8_t pinRx;
+int8_t pinTx;
+int8_t pinEr;
 boolean buttonTriggeredInclusion;
 volatile uint8_t countRx;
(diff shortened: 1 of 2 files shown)


henrikekblad:
This is configurable (by defines in MyConfig.h) in the development branch. And you can add LEDs to sensor nodes. 
Should we still take this in into master you think? Guess it wont hurt, but it will be removed later.

Yveaux:
I ran into issues with my Ethernet shield. It uses pin 4 as Spi ss line,
which was silently also controlled by the gateway for a led.
Do whatever you want with the fix, but others might run into the same
issues. It shouldn't hurt anyways!

BTW. As you know I've been away for a while and I wondered how stable
development branch is at the moment.

Cheers!

Op do 16 jul. 2015 22:16 schreef Henrik Ekblad <notifications@github.com>:

henrikekblad:
The dev branch is stable but need more testers.

Yveaux:
Yeah yeah, I get it ;-)

Op vr 17 jul. 2015 09:13 schreef Henrik Ekblad <notifications@github.com>:

PR #152, opened 2015-08-07
Demoted #error to #warning in SecureActuator.ino

fallberg:


fallberg:
ok, that was not supposed to happen...

fallberg:
Trying again...

PR #174, opened 2015-08-24
Use Tx/Rx pin as IO pin

jkandasa:
With current design we could not use Tx/Rx as IO pin. I have added new variable 'ENABLED_SERIAL'. Disable this if you want to use Tx/Rx as IO pin in atmega328p.


henrikekblad:
Jenkins, build this please.

henrikekblad:
This might be useful. Thanks @jkandasa.

PR #175, opened 2015-08-24
initialize digital pins internal pullup resistor

gizmocuz:
Always send volume when not in sleep mode
send frequency set to 30 seconds

i was receiving interrupts even if i did not connected the pins.
With this patch, it working perfect!

When not in sleep mode, the counter value is now transmitted every 30 seconds, regardless if there is a change.
This way, there will be no timeout


fallberg:
Jenkins, build this please

PR #181, opened 2015-08-27
ESP8266 Test PR

henrikekblad:
For testing the build server


tbowmo:
One question, why is the atmega328 target files changed in this PR request, when it's about ESP8266?

henrikekblad:
What file/line are you talking about?

henrikekblad:
Ahh.. now I understand what you mean.. It is because the IDE build everything in Util.... which result in compile errors on ESP and vice verse.

tbowmo:
ahh.. ok..

fallberg:
Jenkins, build this please.

fallberg:
Jenkins, please build this.........

fallberg:
Jenkins, build this please

fallberg:
Jenkins, build this please.

fallberg:
Jenkins, build this please

libraries/MySensors/MyConfig.h
```diff
@@ -101,7 +101,7 @@
 #define MY_ATSHA204_PIN 17 // A3 - pin where ATSHA204 is attached
 
 // MySigningAtsha204Soft default settings
-#define MY_RANDOMSEED_PIN 7 // A7 - Pin used for random generation (do not connect anything to this)
+#define MY_RANDOMSEED_PIN A0 // A0 - Pin used for random generation (do not connect anything to this)
```
fallberg: I don't like that the defaults are changed for a specific sketch to function. Use constructor arguments in this case.

libraries/MySensors/MyHwATMega328.cpp
```diff
@@ -17,6 +17,8 @@
  * version 2 as published by the Free Software Foundation.
  */
 
+#ifdef __AVR_ATmega328P__
```
fallberg: Is this "ok" for Arduino Mega?

libraries/MySensors/MyHwATMega328.h
```diff
@@ -20,6 +20,8 @@
 #ifndef MyHwATMega328_h
 #define MyHwATMega328_h
 
+#ifdef __AVR_ATmega328P__
```
fallberg: Is this "ok" for Arduino Mega?

henrikekblad:
@Yveaux will have a look on the the problems later tonight.

libraries/MySensors/MyConfig.h
Yveaux: Agree. Only changed this to make experimentation with ESP easier.

libraries/MySensors/MyHwATMega328.cpp
Yveaux: A number of standard #defines are appended to the compiler depending on the target board (you can see it in the commandline behind the e.g. avr-g++ command):

* ESP: ARDUINO_ESP8266_ESP01   ARDUINO_ARCH_ESP8266   ESP8266
* UNO: ARDUINO_AVR_UNO   ARDUINO_ARCH_AVR
* MEGA: ARDUINO_AVR_MEGA2560   ARDUINO_ARCH_AVR
* DUE: ARDUINO_SAM_DUE   ARDUINO_ARCH_SAM  __SAM3X8E__

Then, there are a number or implicit defines introduced by the compiler depending on the target mcu. See e.g. https://gcc.gnu.org/onlinedocs/gcc/AVR-Options.html

I think, to separate between architectures (AVR/ESP/SAM) we can do with the ARDUINO_ARCH_xxx defines.
For this case (and the next one) the #ifdef should be changed to:
#ifdef ARDUINO_ARCH_AVR

**BUT** I find it strange (and that's how this #ifdef sneaked in) that a file called 'MyHw**ATMega328**.cpp' is required by a Mega2560 board.
If this file is indeed shared by all AVR based MCU's, then it should better be called 'MyHwAvr.cpp' or something like it.

fallberg: I agree to that, and my comment is just to give a hint for this error:
http://ci.mysensors.org/job/ArduinoBoardBuilder/a_tool=%2Fopt%2Farduino-1.6.5%2Farduino-headless,arch=avr,board=mega,example=HumiditySensor%2FHumiditySensor.ino,freq=16000000,mcu=atmega2560,pkg=arduino/335/consoleFull#-1860515436253c20e6-3b18-4441-835e-d8f3c24603e1

libraries/MySensors/MyConfig.h
Yveaux: The random generator used by the software SHA204 implementation claims the only analog input pin of the ESP. Can we make use of a different random generator for the ESP implementation?

Yveaux:
I also had to get rid of PinChangeInt library (AVR only) for ESP. In ESP Gateway I simply used standard Arduino attachinterrupt() function. Why is PinChangeInt library used in MySensors?

Same problem with DigitalIO library -- it's AVR only.

We should select carefully on portability when including libraries in the future.

fallberg:
What parts of the atsha implementation is AVR specific?

henrikekblad:
@Yveaux 
Yes, it is unfortunate that master branch still uses PinChangeInt. @thozza managed to remove this dependency in the gateway-refactoring branch by polling the pin instead.  

https://github.com/mysensors/Arduino/tree/gateway-refactoring/libraries/MySensors/examples/SerialGateway

But bringing this into master is quite hard at the moment.

Yveaux:
@henrikekblad I understand there's legacy in the code base, but I think this case can simply be rewritten using standard Arduino api. All Arduino ports will work out of the box with it. 
Anyway, one for the to-do list then!

Yveaux:
@fallberg Thought the ATSHA204.cpp was AVR dependent. Saw some DDR stuff in it and gave up on it. Should it be portable then?

fallberg:
@Yveaux I don't know. What compilation issue did you get?

Yveaux:
Not sure. I'll try to replay the situation tonight.

Yveaux:
Updated the use of the architecture #defines in my repo to be able to compile for Mega again.
Please update from there

Yveaux:
When commenting out the #if defined(ARDUINO_ARCH_AVR) in ATSHA204.cpp I get following errors:

`C:\Users\Ivo\AppData\Local\Temp\build3247091383794380616.tmp\MySensors\utility\ATSHA204.cpp.o 
c:\Users\Ivo\Documents\GitHub\Arduino\libraries\MySensors\utility\ATSHA204.cpp: In constructor 'ATSHA204Class::ATSHA204Class(uint8_t)':
c:\Users\Ivo\Documents\GitHub\Arduino\libraries\MySensors\utility\ATSHA204.cpp:16:18: error: cannot convert 'volatile uint32_t* {aka volatile unsigned int*}' to 'volatile uint8_t* {aka volatile unsigned char*}' in assignment
  device_port_DDR = portModeRegister(port);
                  ^
c:\Users\Ivo\Documents\GitHub\Arduino\libraries\MySensors\utility\ATSHA204.cpp:18:18: error: cannot convert 'volatile uint32_t* {aka volatile unsigned int*}' to 'volatile uint8_t* {aka volatile unsigned char*}' in assignment
  device_port_OUT = portOutputRegister(port);
                  ^
c:\Users\Ivo\Documents\GitHub\Arduino\libraries\MySensors\utility\ATSHA204.cpp:20:17: error: cannot convert 'volatile uint32_t* {aka volatile unsigned int*}' to 'volatile uint8_t* {aka volatile unsigned char*}' in assignment
  device_port_IN = portInputRegister(port);
                 ^
Fout bij compileren.
`

Doesn't seem AVR related, but I didn't invest any time yet to fix it.

fallberg:
Well, I see that the file includes "Arduino.h" so I suppose that makes it totally Arduino dependent. The pin management functions is probably the problem. Is there anything corresponding available that works for both "arduino" and ESP8266?

Yveaux:
Dug a little bit deeper. Aparently these portxxxRegister functions are part of the Arduino API.
AVR has 8 bit port registers, for ESP these macro's are defined in %user%\AppData\Roaming\Arduino15\packages\esp8266\hardware\esp8266\1.6.5-947-g39819f0\cores\esp8266\:

#define portOutputRegister(port)    ((volatile uint32_t*) GPO)
#define portInputRegister(port)     ((volatile uint32_t*) GPI)
#define portModeRegister(port)      ((volatile uint32_t*) GPE) 

ESP has 32bit registers. This explains the uint32_t* vs uint8_t* conflicts.

These portxxxRegister macro's are highly architecture specific, so I wonder if Arduino supports them for e.g. Arm architecture.

The SHA204 implementation just bit-bangs the communication using GPIO's, so it would work just as well when using pinMode/digitalWrite/digitalRead functions.

I personally would prefer rewriting the driver to use the pinMode/digitalWrite/digitalRead functions, or keep it AVR-only for now.

fallberg:
I agree. Lets put an issue on it to be made portable and make it avr only for now.

fallberg:
Perhaps a ugly quick hack would be to do
#ifdef ARDUINO_ARCH_ESP8266
	uint32_t port = digitalPinToPort(pin);	// temoporarily used to get the next three registers
#else //AVR and other 8bit platforms
	uint8_t port = digitalPinToPort(pin);	// temoporarily used to get the next three registers
#endif 
I don't have ESP HW to test on for ATSHA HW I am afraid, and if possible I would prefer not to make too much changes to the driver.
If you do have the HW to test, perhaps you could give it a stab @Yveaux ? But the way I see it, it does not need to be included in this PR unless you really want to. I have created #184 for this.

Yveaux:
I only have an ESP, no SHA.
Let's just make it AVR only for now.
Furthermore it could very well be that the ESP has sufficient CPU power to do all the encryption using the software implementation, but we would have to test that.

fallberg:
Right. I can test it myself (but I have no ESP HW) just to see that it compiles once your PR is in and change it if it works.
Regarding CPU power, in my paranoid world that makes no difference. HW ATSHA is more secure than SW ATSHA.

Yveaux:
I didn't know the software implementation is less secure than the hardware solution.
So, for you to use the ESP gateway with encryption it must be possible to connect an ATSHA to the ESP. Clear.
Anyway, let keep it as is for now, and have the PR address the issue.

So, what's left?
I didn't add support for sleeping the ESP, as the code is meant for a gateway (no need to sleep a gateway and using an ESP as a battery powered sensor is out of scope) and we should keep the ESP running to prevent watchdog timeouts.

Anything more, before we can merge it to master?

fallberg:
Well, technically we are talking signing, not encryption for ATSHA. And it is less secure but the solution is compatible. What I mean is that the quality of the random number generation is not as good in SW as in HW, and the protection of the HMAC key is just as protected as the device is protected from FW readout if anyone would want to determine the key. In HW it is readout protected by design. In SW it is depending on fuses (if it is supported by the device at all to prevent readout) and a question of how secure that protection really is.
And Jenkins, build this please.

libraries/MySensors/utility/RF24.cpp
```diff
@@ -546,22 +546,31 @@ void RF24::startListening(void)
 }
 
 /****************************************************************************/
-static const uint8_t child_pipe_enable[] PROGMEM =
+static uint8_t get_child_pipe_mask(const uint8_t idx)
```
henrikekblad: idx seems to create warnings. 

Should probably just be a fn(void) function.

fallberg:
I made a few changes to the failure cause management that hopefully will provide better details of found warnings in the direct page linked by Jenkins when it finds warnings (the "Details" link under "All checks have failed" and in the link in the comment mysensors-jenkins makes when warnings are found). Jenkins, build this please again so we can see if it worked or not.

Yveaux:
Jenkins, build this please.

libraries/MySensors/utility/RF24.cpp
Yveaux: No, just a plain bug...
Fixed.

Yveaux:
Jenkins, do you only obey @henrikekblad or @fallberg ?
Jenkins, build this please.

fallberg:
Jenkins, build this please

Yveaux:
Thanks @henrikekblad and @fallberg for the great teamwork!

fallberg:
It's what MySensors is all about :)

henrikekblad:
Thank you @Yveaux  for taking the time to let Jenkins getting used to the ESP ;)

Yveaux:
@henrikekblad and @fallberg can we start another Jenkins build or do we need to create a pull-request first?

fallberg:
Jenkins, build this please

fallberg:

Yveaux:
That didn't work, I suppose.
Please refer to https://github.com/mysensors/Arduino/pull/214

PR #189, opened 2015-09-04
Hopefully the last test 

fallberg:
Notable changes compared to official track:
* Management of SPI_EN signal
* Different static IP
* Different MAC
* ATSHA signing support without requests for signatures


fallberg:
A bit too optimistic

PR #191, opened 2015-09-04
A different test

fallberg:


fallberg:
Apparently jenkins believes that this commit has already been built at some point. Cleaver bastard. Illustrerad create one from scratch later to test this throughly

PR #194, opened 2015-09-04
test against dev (should not fail due to PR target)

fallberg:
Notable changes compared to official track:
* Management of SPI_EN signal
* Different static IP
* Different MAC
* ATSHA signing support without requests for signatures


fallberg:
Thanks, you at least started to build. That was all I needed.

PR #196, opened 2015-09-05
Added V_STATUS to S_HEATER, and S_MOISTURE

gizmocuz:
Added V_STATUS to S_HEATER, as this is also supported
Added S_MOISTURE (a moisture sensor)


fallberg:

PR #206, opened 2015-09-15
Added -std=gnu11 compiler flags to MySensor boards

fallberg:
This will allow those boards to be built with the latest Arduino IDE.


fallberg:
Might be ok, but not a needed thing right now.

PR #207, opened 2015-09-22
Signing of messages to gateway

fleinze:
With this change, all messages sent by signing nodes to a gateway that does not require signing are signed.


fallberg:
This is a contradictory change to how the gateway currently works:
The gateway if configured for signing, will require signed messages from all nodes that it signes messages to.
That functionality should cover the usecase you look for. If it does not work, we should fix that, and avoid changing the way signing is supposed to work (and how it is currently documented to work).
I put some more comments on the forum regarding things to try to see if it works:
http://forum.mysensors.org/topic/1021/security-introducing-signing-support-to-mysensors/165?page=9

PR #208, opened 2015-09-22
Mixed signing 2

fleinze:
to match the description in http://forum.mysensors.org/topic/1021/security-introducing-signing-support-to-mysensors :
> However, the difference is that the gateway will only require signed messages from nodes it knows in turn require signed messages.

### DIFF
--- a/libraries/MySensors/MySensor.cpp
+++ b/libraries/MySensors/MySensor.cpp
@@ -567,7 +567,7 @@ boolean MySensor::process() {
 #ifdef MY_SIGNING_FEATURE
 	// Before processing message, reject unsigned messages if signing is required and check signature (if it is signed and addressed to us)
 	// Note that we do not care at all about any signature found if we do not require signing, nor do we care about ACKs (they are never signed)
-	if (signer.requestSignatures() && msg.destination == nc.nodeId && mGetLength(msg) && !mGetAck(msg) &&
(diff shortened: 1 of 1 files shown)


fallberg:
@fleinze Have you tested this change with a GW that require signatures, and two nodes (on that is "secure" and one that is "insecure")?
Expected behavior:
GW<->"secure node": all messages are signed in both directions
GW<->"insecure node": no messages are signed at all

I cannot currently verify the change myself I am afraid. But this is at least how it is supposed to work.

Also, you should be able to set the GW to not require signatures, then the behavior should be like this:
GW<->"secure node": all messages to the node is signed, no messages to the GW is signed
GW<->"insecure node": no messages are signed at all
It would be great if you could test this usecase as well.

fallberg:
Jenkins, build this please

fleinze:
thank you fallberg, I did test the usecases:
"insecure" GW -> "secure" node: messages to gateway are not signed, messages to node are signed
"insecure" GW -> "insecure" node: no signing at all

fallberg:
Ok, so all testcases did pass as expected?

fleinze:
I just did final tests with a clean copy: (true means requires signing, false means doesn't require signing, as in the MySigning constructor)
GW(true) -> node(true): all messages signed
GW(true) -> node(false): no messages signed
GW(false) -> node(false): no messages signed
GW(false) -> node(true): messages to the gw are not signed, messages to the node are signed
So, all testcases as expected.

fallberg:
Thanks @fleinze for finding and fixing this!

PR #210, opened 2015-09-24
Forward declare all inline functions

fallberg:
This seem to be mandatory for newer Arduino IDE versions.


fallberg:
Jenkins, build this please. And try to not gobble all memory.

PR #214, opened 2015-10-03
Debug output fixes for ESP8266

Yveaux:
Fixed printDetails() to work for ESP8266.
Replaced printf_P calls by Serial.print, as printf_P tends to crash on ESP8266.
Simplified debugging of driver by only #defining DEBUG (removed MINIMAL & SERIAL_DEBUG).


henrikekblad:
Jenkins, add author to whitelist please.

PR #221, opened 2015-10-06
Updated the MySensors internal SPI flashdriver

fallberg:
Now matches the "public" SPI flashdriver better.


fallberg:
That was not good enough. The drivers need to have unique namespaces or the nightly version of the IDE will consider them duplicates.

PR #235, opened 2015-10-28
Changed library version to comply with Semantic Versioning

fallberg:
New version: 1.6.0-beta for development of 1.6.0.
See http://semver.org

### DIFF
--- a/libraries/MySensors/core/Version.h
+++ b/libraries/MySensors/core/Version.h
@@ -6,6 +6,6 @@
 #ifndef Version_h
 #define Version_h
 
-#define LIBRARY_VERSION "1.6"
+#define LIBRARY_VERSION "1.6.0-beta"
 
 #endif


fallberg:
This is considered a PROPOSAL. @henrikekblad may merge if he consider this a good way forward.
Else, lets discuss :)

bblacey:
Semantic versioning is very common place so it gets my vote (we use it at work for all our products).  Would be wise to script the exact MySensors convention and ideally have the build system maintain it to avoid human error and we should take it to maximum precision (i.e. finer resolution than -beta by either using the commit short-sha or some prerelease convention that represents a git tag).

fallberg:
Yes, but it is a bit tricky to automate the versioning vs backwards compability. But if you have any clue on how to automate this, feel free to go ahead :)

PR #240, opened 2015-11-03
HeatpumpIR: Support Sharp/IVT heatpumps

ToniA:
* Support Sharp/IVT
* Change the IR carrier frequency of all heatpumps to 38 kHz
* 'Samsungbyte' was accidentally changed to 'Samsunguint8_t' when replacing 'byte' with 'uint8_t'


henrikekblad:
Thanks for the update.

PR #243, opened 2015-11-07
Add V_CUSTOM for controller/inter-node specific commands.

henrikekblad:
Requested by rob@domoticz for handling request of internal device
values. This variable/command should not be used for any of the official
sketches among the examples when communicating with controller.


fallberg:
No risk this becoming a "slaskhink" for various messages that better could have their own type?

henrikekblad:
Yep, already warned Rob about this. Any commonly cross controller usable vars should have their specific well documented vars.  

In the domoticz case it will be used for requesting data from none-mysesnors devices. Right now it's hard to make a common least significant "divisor" for this functionality.

PR #245, opened 2015-11-08
implementation of I_DISCOVER

tekka007:
Discovery function for all active (=listening) nodes. Nodes reply with
I_DISCOVER_RESPONSE and the parent node as payload. Useful to retrieve
network topology.


henrikekblad:
Cool, did you get a chance to test this through a repeater.

tekka007:
Yes, repeater does controlled forwarding

henrikekblad:
Awesome. I'll merge it. 

Maybe 1.6 nodes should send a I_CAPABILITIES message instead of I_CONFIG at startup with the extended information... or both to be backward compatible.

tekka007:
yes, there are still things tbd in terms of included information and encoding (i.e. bit vs. comma-separated values)...

PR #259, opened 2015-11-19
Fixed bug in repeaters when forwarding signed messages

fallberg:
For repeated messages, sign flag must not be touched.


fallberg:
Any opinions fellow moderators?

fallberg:
Patch has been tested by forum member who reported the issue (mannkind) but maybe there are other architectural concerns?

PR #260, opened 2015-11-21
I_HARDWARE_CONFIG

tekka007:
new internal message to report HW parameters such as freq, fuse
settings, etc.


tekka007:
yes - both should be excluded from signing: I_DISCOVER is a broadcasting message to discover the topology of listening nodes w/o need for signing. I_HARDWARE_CONFIG is sent from the bootloader to report various HW parameters. This function cannot be placed in the framework unless the bootloader would forward or store (in eeprom) parameters like MCU status register (for the reset cause) and others.

fallberg:
Ok, right. Is it you or do you know who, will implement these? Just so that the relevant changes are made also in the signing logic in the library. I believe the I_PRESENTATION handling can be used as reference (that is also an example of a message that is not signed).

fallberg:
Will you add the code to handle exclustion of I_DISCOVER and I_HARDWARE_CONFIG so that they are not picked up by the signing logic? You can use I_ID_REQUEST as an example (only internal messages are currently among the ones that are exempt from signing).

henrikekblad:
Closing this after our discussion today @tekka.

PR #264, opened 2015-11-23
esp8266gateway crashing when receiving I_CONFIG message from new node…

soward:
… due to attempt to dereference the NULL return of MyMessage::getString()

this fix checks for NULL return from getString and defaults isMetric to true, otherwise attempts the usual match. Works for me so far. Not tested on AVR.

( I AGREE TO THE CLA )

### DIFF
--- a/libraries/MySensors/core/MySensorCore.cpp
+++ b/libraries/MySensors/core/MySensorCore.cpp
@@ -241,7 +241,11 @@ void _processInternalMessages() {
 	} else if (type == I_CONFIG) {
 		// Pick up configuration from controller (currently only metric/imperial)
 		// and store it in eeprom if changed
-		isMetric = _msg.getString()[0] == 'M' ;
+	 	if (_msg.getString() == NULL) {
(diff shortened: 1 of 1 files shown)


henrikekblad:

henrikekblad:
@soward  
Thanks for spotting this. Just wonder why crash doesn't happen for everybody.

soward:
Not sure exactly, maybe depends on the exact 8266 device used? This page suggests that 0 and 1 are protected and might be cleared for just such purposes:

http://esp8266-re.foogod.com/wiki/Memory_Map <http://esp8266-re.foogod.com/wiki/Memory_Map>

but in the crash output always had the same address in the high 0x4010 range, perhaps to some flash that doesn’t exist on the board I was using? ( nodemcu v 0.9 ‘yellow’ board )

bblacey:
FYI, I experienced a similar crash when experimenting with a NodeMCU 1.0...  It happened when I rebooted one of my nodes...  Good catch!

PR #269, opened 2015-11-28
Fix DallasTemperatureSensor Example

51systems:
presentation() is called before setup(), so unless we call sensors.begin() in presentation(), sensors.getDeviceCount() will always return zero.

### DIFF
--- a/libraries/MySensors/examples/DallasTemperatureSensor/DallasTemperatureSensor.ino
+++ b/libraries/MySensors/examples/DallasTemperatureSensor/DallasTemperatureSensor.ino
@@ -53,13 +53,14 @@ MyMessage msg(0,V_TEMP);
 
 void setup()  
 { 
+}
+
+void presentation() {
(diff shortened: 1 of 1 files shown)


bblacey:
Jenkins, build this please.

tbowmo:
Perhaps the core should be changed instead, so presentation is called after setup.
By including this patch, you will reinitialize the one wire network everytime it receives a request for presentation.

henrikekblad:
Yes, agree with @tbowmo here, better if we call sketch setup before presentation.
I'll push something. 

Maybe you could verify it later @51systems?

henrikekblad:
Ok, fix pushed.

@51systems , can you try again with the original sketch?

51systems:
Yep, that did the trick.

I really like the new configuration paradigm you're using for 1.6! It looks like its going to give a lot more flexibility while hiding the boilerplate. 

Many thanks

PR #301, opened 2015-12-16
Resolve open bugs relating to security

fallberg:
I have not the opportunity to verify these at the moment. If someone has that possibility within the next few days, I would highly appreciate some help (ping @tekka007 ;) ).


tekka007:
Verified, together with #302

PR #303, opened 2015-12-17
Adding SI7021 based TempNode example in 1.6 format

GertSanders:
Adding SI7021 based TemNode example in 1.6 format.
Adding Vcc library.


tbowmo:
Perhaps it's just me, but isn't that almost the same functionality as the Sensebender sketch?

GertSanders:
Maybe we just need an empty sketch template with all possible defines commented out and anotated so their function becomes clear from the accompanying comments.

People switching from 1.5 to 1.6 seem to ask the same questions I resolved in my sketch by adding commented defines so I can use them in the future.

Gert

GertSanders:
Will use the sketch for personal projects only. Pull request 304 took care of code for Vcc measuring in battery based setups. This pull request can be closed.

PR #304, opened 2015-12-17
Core functions for Vcc and frequency readings

tekka007:
Implementation of functions to measure
- hwCPUVcc() in mV
- hwCPUFrequency() in Mhz


tekka007:
with this PR we can omit the Vcc part of #303.

PR #308, opened 2015-12-24
@PlatformIO Library Registry manifest file

ivankravets:
* This library in Web Registry: http://platformio.org/#!/lib/show/548/MySensors
* Specification: [library.json](http://docs.platformio.org/en/latest/librarymanager/config.html)
* Library Manager:  http://docs.platformio.org/en/latest/librarymanager/index.html


henrikekblad:
Thanks Ivan!

ivankravets:
@henrikekblad thanks you too for the merge. Happy Holidays and Merry Christmas!

PR #310, opened 2015-12-26
Fix wrong value returned in isAckPayloadAvailable()

tekka007:
Logical not (!) precedes bitwise and (&), thus wrong evaluation of the return term.


henrikekblad:
Wot.. If this is wrong you should probably notify @TMRh20 about it.

fallberg:
The new behaviour sure would make more sense than the old one.

tekka007:
Aha, TMRh20 fixed his library. Bug was also reported by @jorihardman: https://github.com/TMRh20/RF24/pull/83

PR #314, opened 2015-12-28
ClearEepromConfig: Use EEPROM.update instead of EEPROM.write

mfalkvidd:
EEprom has limited write cycles, so don't write unless we need to.
This also speeds up the clearing. My test results:
write: 1,700ms
update, all cells needs to be updated: 1,600ms
update, no cells need to be updated: 4ms

See https://www.arduino.cc/en/Reference/EEPROMUpdate for more
information on the udpate function.

I agree to CLA.

### DIFF
--- a/libraries/MySensors/examples/ClearEepromConfig/ClearEepromConfig.ino
+++ b/libraries/MySensors/examples/ClearEepromConfig/ClearEepromConfig.ino
@@ -32,7 +32,7 @@ void setup()
   Serial.begin(MY_BAUD_RATE);
   Serial.println("Started clearing. Please wait...");
   for (int i=0;i<512;i++) {
-    EEPROM.write(i, 0xff);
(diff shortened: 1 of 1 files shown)


fallberg:
Jenkins, build this please

tekka007:
It would be better to implement a hardware independent version to clear the EEPROM. Maybe you could have a look at the core HwXXXXXX drivers and use the hwWriteConfig() function instead (which also uses the update function).

mfalkvidd:
Sorry. I was unaware that EEPROM.update was less portable than EEPROM.write. I have no knowledge of HwXXXXXX so I am unable to assist. Feel free to deny this pull request.

tekka007:
Merged; intermediate solution.

fallberg:
Which commit implements intermediate solution?

PR #323, opened 2016-01-04
LED HW definitions in arduino HW files overridable

tbowmo:
Use MY_HW_xx_LED_PIN, if it is defined in board specific files, AND if MY_DEFAULT_xx_LED_PIN is not defined.

Commented out MY_DEFAULT_xx_LED_PIN defines in gateway sketches

if MY_DEFAULT_XX_LED_PIN is defined, it also enables MY_LEDS_BLINKING_FEATURE

If MY_LED_BLINKING_FEATURE is enabled, but no MY_DEFAULT_xx_LED_PIN are defined, then it will undefine MY_LED_BLINKING_FEATURE, as it doesn't make sense to include it if no LED's are defined.

### DIFF
--- a/hardware/MySensors/samd/variants/mysensors_gw/variant.h
+++ b/hardware/MySensors/samd/variants/mysensors_gw/variant.h
@@ -191,9 +191,9 @@ static const uint8_t SCK  = PIN_SPI_SCK ;
 
 
 #define MY_SIGNING_ATSHA204_PIN 17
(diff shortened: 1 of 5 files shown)


fallberg:
Good to go?

tbowmo:
Fine by me. Just rebased against development.

tor. 7. jan. 2016 22.05 skrev Patrick Fallberg <notifications@github.com>:

PR #325, opened 2016-01-05
HW-independent ClearEepromConfig, substitutes #314

tekka007:


fallberg:
Perhaps add MY_CORE_ONLY as a comment in MyConfig.h so it becomes somewhat documented and visible? (doxygen is preferred but we can also do a major doxygenification of myconfig at a later stage since there are a lot of undocumented flags there(.

PR #328, opened 2016-01-06
Fix MY_WITH_LEDS_BLINKING_INVERSE always on/blink off

sundberg84:
Correct MY_WITH_LEDS_BLINKING_INVERSE (Include now means leds are always lit)

### DIFF
--- a/libraries/MySensors/core/MyLeds.h
+++ b/libraries/MySensors/core/MyLeds.h
@@ -22,11 +22,11 @@
 
 
 #ifdef MY_WITH_LEDS_BLINKING_INVERSE
-#define LED_ON 0x1
-#define LED_OFF 0x0
-#else
 #define LED_ON 0x0
 #define LED_OFF 0x1
+#else
+#define LED_ON 0x1
+#define LED_OFF 0x0
 #endif
 
 #ifdef MY_LEDS_BLINKING_FEATURE


fallberg:
Please be more descriptive in the commit headline. The headline gives no information on what you actually changed.

sundberg84:
Sorry, first request... also didnt sign any agreement but whatever that is i agree to that.

fallberg:
Jenkins, build this please

fallberg:
Actually, your commit headline remains unchanged. Please remember to format your commit message a bit better next time :) 
Here is a good guide: http://chris.beams.io/posts/git-commit/

sundberg84:
It changed here in the header on the pull request but not in the build :/
Yes, will read - thank you for the link.

fallberg:
You can't change a commit message without amending the commit and pushing the updated commit. The pr header is the label the merge commit will get but tour own commit will also show in the git history, hence the need for well written commit messages so the history makes sense.

fallberg:
Regarding the actual change, isn't this depending on how the led is connected? What is the default for MySensors? To sink the led or source it? @henrikekblad do you know?

sundberg84:
I have it connected 7,8,9 to Gnd on a Uno with Ethernet Shield if you need that info.

fallberg:
Ok, because the official building instructions have them connected to VCC and the pin sinks the net. So the default behaviour is correct with those instructions.

sundberg84:
Oh, sorry - my misstake then... :( I must have missed that...
Then we can just close/remove this pull I guess.

PR #330, opened 2016-01-07
Add support for the CC3000 wifi module to be used as a ethernet gateway.

pressy4pie:


henrikekblad:
Hmm.. the example seems to have been build against master branch. I don't think it will compile... 

But lets's see what Jenkins says...

Jenkins, build this please.

pressy4pie:
oops your right it was on master branch. sorry about that.

PR #342, opened 2016-01-13
Use the internal pullup to be able to hook up this sketch directly to S0 meters

marcovannoord:
If no pullup is used, the reported usage will be too high because of the floating pin.

Since an S0-enabled meter acts like a switch, we need a resistor to keep the pin from floating.


fallberg:
Consider using pinMode to configure internal pullup: https://www.arduino.cc/en/Reference/PinMode
That makes the code easier to read since it is a more obvious way of enabling internal pullup. It ought also be more portable for other architectures supporting the arduino api:s.

marcovannoord:
Thanks, i didn't know that last part, i've edited my code to resemble your comment

fallberg:
Nice :) 
Jenkins, build this please

PR #344, opened 2016-01-14
Update RF24.cpp

Kolbi:


fallberg:
I notice a lot of pull requests and all commit messages just indicate the file that was updated. It would be better if the commit message stated what was changed. Then one could just read the git log in condensed form to get an overview on what has happened. This is a great guide to writing good commit messages: http://chris.beams.io/posts/git-commit

Existing pr:s can be updated by amending the commit and rewrite the message (git commit --amend) and then force pushing the branch to github (git checkout ;git push origin HEAD: -f)

PR #351, opened 2016-01-16
Adding SMS Gateway node sketch in Examples

GertSanders:
Removed my local copy of Adafruit FONA from MySensors libraries.
Added comments in sketch to inform user that Adafruit FONA library must
be added via Library Manager in Arduino IDE to compile this sketch.

Added the SMS Gateway node sketch as an example of using the SIM800L module found on Aliexpress (works worldwide as it is a 4G GSM module).


GertSanders:
So it seems I DO need to add the Adafruit FONA Library after all.

PR #352, opened 2016-01-16
Adding the corrected Adafruit FONA library

GertSanders:
Adding the corrected Adafruit FONA library again (for the last time I
hope).

Builds of the example sketch fail because the FONA library is not a standard library. A user needs to add it via the library manager or it must be in the MySensors set of libs. Either way work, but only adding it in the MySensors library allows getting Jenkins approval.


henrikekblad:
Might wanna initialize than pointer like this:

char* ptr = NULL;

GertSanders:
Back to the gym with Chuck ...

PR #353, opened 2016-01-16
Chuck Norris punched this char array init knockout

GertSanders:
An initialisation of a char array was giving me grief. What baffles me is that it passed some build types (GW) and failed others (nano, pro ...)
Finally the init sequence used now seems to pass compile without any warning. This code is not used in SIM800L because it has no GPS, but anyway, we should be fine now. I hope. Really.


henrikekblad:
Ok, nice no warnings! 
But you must have done something wrong when starting this pr.. You got commits from master branch in here. https://github.com/mysensors/Arduino/pull/353/commits 

Suggest you:

Make a copy of your local repo  so you won't lose them.

* Re-clone the MySensors library here on github.
* Clone it to your computer
* > git checkout development
* copy the new library and example
* commit, push and make a new pr (with a nice commit message)

GertSanders:
Will recline and resubmit

PR #354, opened 2016-01-17
Adding SMS Gateway node example and supporting library.

GertSanders:
Adding SMS Gateway node example and supporting library (Adafruit FONA).


henrikekblad:
Now we're talking!

Thanks!

fallberg:
Gold star

PR #357, opened 2016-01-19
The gateway sent wrong signing information to node with MY_SIGNING_RE…

zaphodus:
…QUEST_SIGNATURES enabled

The gateway does not care about the node's signing state, while msg.data[1] = 0; in the signature verification process.

In this case the part of the log of the gateway:
0;255;3;0;9;Mark node 3 as one that require signed messages
0;255;3;0;9;Mark node 3 as one that do not require whitelisting
0;255;3;0;9;Informing node 0 that we do not require signatures
0;255;3;0;9;Informing node 0 that we do not require whitelisting

I removed this line, and now it works properly.

### DIFF
--- a/libraries/MySensors/core/MySigning.cpp
+++ b/libraries/MySensors/core/MySigning.cpp
@@ -102,7 +102,7 @@ static void prepareSigningPresentation(MyMessage &msg, uint8_t destination) {
 	build(msg, _nc.nodeId, destination, NODE_SENSOR_ID, C_INTERNAL, I_SIGNING_PRESENTATION, false).set("");
 	mSetLength(msg, 2);
 	msg.data[0] = SIGNING_PRESENTATION_VERSION_1;
-	msg.data[1] = 0;
+//	msg.data[1] = 0;
 }
 
 void signerInit(void) {


fallberg:
I don't see the problem in your log. Are you sure you have configured your gateway to require signatures? To me, the log looks perfectly fine for a gateway that so not require signatures.

fallberg:
To break down your log (which is coming from a gw):
The first line indicate that the GW acknowledge that a node require signing.
The second that the node does not use whitelisting. 
The third that the GW does not require signing (I suppose it is not enabled on the GW) 
The fourth that the GW does not use whitelisting.

zaphodus:
Both sides are set to require signatures.

fallberg:
Hm. Ok, I'll like closer on the code. But your proposed fix just fixes the symptom not the problem. The initialization is supposed to look like that. If anything, the proper bit is then not set for a gw that require signatures. Can you confirm that you really do have that enabled on the GW. I have tested the same scenario when I developed the code.

fallberg:
Ok, I think I found the problem you see. It has to do with msg.sender being reset before gw sends it's preferences. Your pr won't really fix that problem though, but I will create a new one for you to test.

fallberg:
And now I see the problem in your log. Sorry, missed that. But it is a good confirmation on my suspicion. The sender is reset. I know what's causing this. I will file a PR to fix it (this commit does not I am afraid)

PR #358, opened 2016-01-19
Make sure signing presentation data is sent correctly

fallberg:
The message sender was retrieved from the message buffer after
prepareSigningPresentation() was called. This function resets
the message buffer so that is not the proper way to do it. Now,
the message sender is extracted first in signerProcessInternal()
in order to avoid the value being overwritten later on.


fallberg:
@zaphodus could you please try this fix. It should resolve your issue.

zaphodus:
I tried this fix and it works well. Thank you.

fallberg:
Great! Thanks for confirming.

PR #359, opened 2016-01-19
Cleaned up Doxygen warnings

fallberg:


fallberg:
It's worth mentioning that this PR only addresses the Doxygen warnings. It does not provide any documentation more than just adding a brief documentatory one-liner to the parts doxygen consider undocumented.

PR #365, opened 2016-01-20
Fix version to stable

ivankravets:


fallberg:
This pr will go into development. It will be misleading to identify 2.0.0-beta as 1.5.3 I believe

ivankravets:
Please take a look at http://docs.platformio.org/en/latest/librarymanager/config.html#version and "Case 1". In your case, the tag/release will be used. See actual library archive http://platformio.org/#!/lib/show/548/MySensors

fallberg:
I see.

PR #377, opened 2016-01-29
Node lock feature

### LINKED ISSUE #280: Implement brute force attack prevention
A node should count the number of subsequent failed signature verifications and when reaching a configured limit, send a message to gw/controller that a brute force attack is being attempted. It should then write a byte to eeprom and reboot. At boot up signing backend checks this byte during early init and decides if full startup is to be permitted else lock down by sleeping/rebooting. Possibly resending a warning to gw/controller. To recover, node needs to be reprogrammed to reset the eeprom setting.

fallberg:
When activated, suspicious signing related activities will eventually
cause the node to lock down and transmit a I_LOCKED message to the
gateway at 30m intervals.
Unlocking is done by grounding a pin or erasing the EEPROM.
See Doxygen for details on use.
Fixes #280


libraries/MySensors/core/MyEepromAddresses.h
```diff
@@ -38,6 +38,7 @@
 #define EEPROM_SIGNING_SOFT_HMAC_KEY_ADDRESS (EEPROM_WHITELIST_REQUIREMENT_TABLE_ADDRESS+32) // This is set with SecurityPersonalizer.ino
 #define EEPROM_SIGNING_SOFT_SERIAL_ADDRESS (EEPROM_SIGNING_SOFT_HMAC_KEY_ADDRESS+32) // This is set with SecurityPersonalizer.ino
 #define EEPROM_RF_ENCRYPTION_AES_KEY_ADDRESS (EEPROM_SIGNING_SOFT_SERIAL_ADDRESS+9) // This is set with SecurityPersonalizer.ino
-#define EEPROM_LOCAL_CONFIG_ADDRESS (EEPROM_RF_ENCRYPTION_AES_KEY_ADDRESS+16) // First free address for sketch static configuration
+#define EEPROM_NODE_LOCK_COUNTER (EEPROM_RF_ENCRYPTION_AES_KEY_ADDRESS+16)
```
henrikekblad: Maybe switch order of these. Will screw peoples local config otherwise.

fallberg: Ooops. That is a merge conflict I did not resolved properly. Will fix

fallberg: But when I look at it again, I see no problem. EEPROM_LOCAL_CONFIG_ADDRESS is on line 42, and last in the array.

henrikekblad: Ok,  I'll create an issue to move EEPROM_LOCAL_CONFIG_ADDRESS starting from end of EEPROM so we can add more configurations without shifting local config in the future.

henrikekblad:
out-of-date again.. Need to find out why is this happening all the time. New "feature" of github or did we change some branch config that affects the behaviour?

libraries/MySensors/core/MyEepromAddresses.h
fallberg: I think it is better the other way around. Users own configs can start from 0,the library internal settings can grow from the end.

PR #385, opened 2016-02-05
Resend presentation if client connected

henrikekblad:
This change solve the problem described  here:
http://forum.mysensors.org/topic/2112/n00b-ethernet-gw-without-nf24

The presentation must be resent when the client connects. Without this Domoticz catches only sensor data, and doesn't know about connected actuators.


zaphodus:
Domoticz: V2.4041
MySensors: 2.0.0-beta development branch 2016-02-13

PR #386, opened 2016-02-06
Add before() for initialisations before radio hogs SPI

### LINKED ISSUE #293: preBegin Function
Hey guys,

at the moment i am working on an sensor node with an MRFC522 RFID reader beside the NRF24 on the SPI Bus. 

With Version 1.5.1 it was possible to run code before the radio init function ( gw.begin(); )so that you could do something like this to stop the rfid reader on the SPI Bus so that there where no interferences on the bus with the NRF Module:
```
  pinMode(RST_PIN, OUTPUT);     
  digitalWrite(RST_PIN, LOW);
  pinMode(SS_PIN, OUTPUT);     
  digitalWrite(SS_PIN, LOW);
  gw.begin();
```

In Version 1.6 you run the setup() function after the radio init in your _beginn function so that above mentioned code comes to late and the radio init fails.

Is it possible that you could implement something like the loop or setup function that comes before? So that we can do such thing before?

I would implement this by myself but my CFu is not very good. 

// Initialize library and handle sketch functions like we want to
```
int main(void) {
	init();
	#if defined(USBCON)
		USBDevice.init();
    		USBDevice.attach();
	#endif
          if (preBegin) preBegin();  <------
	_begin(); // Startup MySensors library

	for(;;) {
		_process();  // Process incoming data
		if (loop) loop(); // Call sketch loop
		if (serialEventRun) serialEventRun();
	}
	return 0;
}
```

Regards 

Florian

henrikekblad:
Fix #293

### DIFF
--- a/libraries/MySensors/core/MySensorCore.cpp
+++ b/libraries/MySensors/core/MySensorCore.cpp
@@ -64,6 +64,9 @@ void _begin() {
 	    hwInit();
 	#endif
 
+	// Call before() in sketch (if it exists)
+	if (before) before();
+
 	debug(PSTR("Starting " MY_NODE_TYPE " (" MY_CAPABILIIES ", " LIBRARY_VERSION ")\n"));
 
 	signerInit();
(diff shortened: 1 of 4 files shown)


fallberg:

PR #391, opened 2016-02-08
Set the serial port speed of the Serial Gateway to 38400 on Arduino's  running on 8MHz oscillator

ToniA:
On 8 MHz clock speed the actual 115200 bps rate is so much off that sending
data over serial is very unreliable.

### DIFF
--- a/libraries/MySensors/examples/GatewaySerial/GatewaySerial.ino
+++ b/libraries/MySensors/examples/GatewaySerial/GatewaySerial.ino
@@ -51,6 +51,11 @@
 // Enable serial gateway
 #define MY_GATEWAY_SERIAL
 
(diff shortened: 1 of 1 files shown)


henrikekblad:
8Mhz should be able to handle 57k6. Could you update?

ToniA:
According to the calculator, the 57600 serial speed is quite much off from 57600. 38400 is almost a perfect match, the error is only 0.2%. And I also managed to have 38400 as the other speed option in Domoticz (it has a drop-down list with two speeds now).

http://wormfood.net/avrbaudcalc.php?bitrate=300%2C600%2C1200%2C2400%2C4800%2C9600%2C14.4k%2C19.2k%2C28.8k%2C38.4k%2C57.6k%2C76.8k%2C115.2k%2C230.4k%2C250k%2C.5m%2C1m&clock=8&databits=8

PR #392, opened 2016-02-08
Major RF24 library rework: minimal driver at full specs

tekka007:
- strip C++ code
- 2-fold speed increase
- reduce code size ~2kb
- eliminate non-essential register polling/writing
- use 2 RX pipes


libraries/MySensors/MyConfig.h
```diff
@@ -35,11 +35,6 @@
 // final sketch but is helpful to see what is actually is happening during development
 //#define MY_DEBUG
 
-// Enable MY_DEBUG_VERBOSE flag for verbose debug prints related to RF24 radio.
-// Requires DEBUG to be enabled.
-// This will add even more to the size of the final sketch!
-//#define MY_DEBUG_VERBOSE
-
```
tbowmo: Why is this removed from MyConfig.h?

henrikekblad: Seems to be called MY_DEBUG_VERBOSE_RF24 now. A more specific define is fine by me.  We should perhaps keep an updated comment in MyConfig so it's easier to find.

fallberg: I suggest adding a commented section, like MY_DEBUG_VERBOSE_SIGNING below. Perhaps even add doxygen tags to these so that the debug features gets documented. But that can be made in a separate commit. I encourage everyone to improve on the doxygen parts :)

tekka007:
jenkins, build examples please

tekka007:

tekka007:

tekka007:

tekka007:

PR #400, opened 2016-02-17
Allow sending multiple commands in one serial transmission

henrikekblad:
E.g. 
"1;2;1;0;2;1\n1;11;1;0;2;1\n1;1;1;0;2;0\n1;2;1;0;2;0\n"

### DIFF
--- a/libraries/MySensors/core/MyGatewayTransportSerial.cpp
+++ b/libraries/MySensors/core/MyGatewayTransportSerial.cpp
@@ -43,18 +43,18 @@ bool gatewayTransportInit() {
 
 
 bool gatewayTransportAvailable() {
-	bool available = false;
-
 	while (MY_SERIALDEVICE.available()) {
 		// get the new byte:
 		char inChar = (char) MY_SERIALDEVICE.read();
 		// if the incoming character is a newline, set a flag
 		// so the main loop can do something about it:
-		if (_serialInputPos < MY_GATEWAY_MAX_RECEIVE_LENGTH - 1 && !available) {
+		if (_serialInputPos < MY_GATEWAY_MAX_RECEIVE_LENGTH - 1) {
 			if (inChar == '\n') {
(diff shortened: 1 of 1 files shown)


maratkh:
Where i can see release for this feature? I need to revive multiple command in one message. What possible max lengt for one message (multiple command message)&

roboprint:
Is it correct that this code will not be executed if you returning true after finding first '\n'?
```
	if (available) {
		// Parse message and return parse result
		available = protocolParse(_serialMsg, _serialInputString);
		_serialInputPos = 0;
	}
```

It seems that it would be more correct to call `protocolParse()` before returning true or `gatewayTransportReceive()` call in gateway code will return previous parsed message.

henrikekblad:
Very true! Good catch. I'll update.

TEAM: NO MERGING OF THIS YET.

maratkh:
Greate! We have already encountered this opportunity yesterday. Packets sent to the node "cut". We will try this thix tomorrow

roboprint:
@henrikekblad sorry, but seems like your fixed version contains logic error. Protocol will hang up on first invalid message (for example, if message parameters count is < 5).

In this case `protocolParse()` will return `false` and `_serialInputPos` variable will not be set to 0, so next received `'\n'` will cause to return false from `protocolParse()` again until buffer is not overflowed (MY_GATEWAY_MAX_RECEIVE_LENGTH).

henrikekblad:
Thanks, just can't seem to get things right today ;)

JohanElmis:
Hello, 

I  know you're gonna be surprised  by that  news,  just take  a look <http://lap.fontanossubschicago.com/e4ejysl>

Hugs, johan

JohanElmis:
Dear, 


I've recieved a message from our friends,  and they  have a surprise  for  us!  Just take  a look <http://phibyngede.dougshrader.com/e4ugxwk>

johan

PR #414, opened 2016-03-25
Added WasherMonitor to MySensors Development and i2c LCD Library

drock1985:
This is the code for the MySensors Contest2016. It reads the LED status of a washing machine and reports back the combination of lights on/off back to in plain text the status of the washing machine!


henrikekblad:
This has already been added to https:///openhardware.io which is the right place.

We will move the more exotic examples there in the future.

PR #419, opened 2016-04-09
Fix exception in esp8266 gateway. 

gerritv:
An incompatibility exists with the wifi SetNoDelay() code between esp8266/Arduino version 2.0.0 and 2.1.0. This patch solves the problem for MySensors ESP8266 gateway until the SDK is fixed.
See https://github.com/esp8266/Arduino/issues/1695 for details of the actual problem.
The symptom is a gateway that keeps restarting after connection to controller due to an exception.

The solution was found by @sle118 in Mar 2016


tekka007:
Thanks for reporting this - surprisingly, there is no issue with 2.1.0-rc2.

PR #420, opened 2016-04-09
Present the gateway itself before gateway child sensors

davidedmundson:
This fixes child nodes working under home-assistant

### DIFF
--- a/libraries/MySensors/core/MyTransport.cpp
+++ b/libraries/MySensors/core/MyTransport.cpp
@@ -452,7 +452,10 @@ void transportPresentNode() {
 	// Open reading pipe for messages directed to this node (set write pipe to same)
 	transportSetAddress(_nc.nodeId);
 	// Present node and request config
-	#ifndef MY_GATEWAY_FEATURE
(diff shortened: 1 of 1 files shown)


henrikekblad:
Could this be fixed in HomeAssistant instead? A bit worried how other controllers would react if we introduced this message.

davidedmundson:
Other controllers will need it too.

Gateways having child sensors is new. 
Its mysensors "breaking" behavior by now having sensors without a presented sensor.

This patch brings the serial API to be the same as previous releases from the POV of the controller.

henrikekblad:
Hmm, true. It should probably be defined as a repeater anyhow. Could you update this?

henrikekblad:
When thinking a bit more it should probably be:

#if defined(MY_REPEATER_FEATURE)
   present(NODE_SENSOR_ID, S_ARDUINO_REPEATER);
#else
   present(NODE_SENSOR_ID, S_ARDUINO_NODE);
#endif

henrikekblad:
Fixed in #426

PR #422, opened 2016-04-14
Accelerate getnonce with single ATSHA random command, frame ATSHA cal…

paukstis:
…ls with wakeup-sleep/idle to avoid spending time in sha204c_resync

### DIFF
--- a/libraries/MySensors/core/MySigningAtsha204.cpp
+++ b/libraries/MySensors/core/MySigningAtsha204.cpp
@@ -107,15 +107,14 @@ bool signerAtsha204GetNonce(MyMessage &msg) {
 	// Generate random number for use as nonce
 	// We used a basic whitening technique that takes the first byte of a new random value and builds up a 32-byte random value
 	// This 32-byte random value is then hashed (SHA256) to produce the resulting nonce
-	for (int i = 0; i < 32; i++) {
-		if (atsha204.sha204m_execute(SHA204_RANDOM, RANDOM_NO_SEED_UPDATE, 0, 0, NULL,
-											RANDOM_COUNT, _singning_tx_buffer, RANDOM_RSP_SIZE, _singning_rx_buffer) != SHA204_SUCCESS) {
-			DEBUG_SIGNING_PRINTBUF(F("Failed to generate nonce"), NULL, 0);
-			return false;
-		}
-		_signing_current_nonce[i] = _singning_rx_buffer[SHA204_BUFFER_POS_DATA];
+	atsha204.sha204c_wakeup(_signing_temp_message);
+	if (atsha204.sha204m_execute(SHA204_RANDOM, RANDOM_NO_SEED_UPDATE, 0, 0, NULL,
+										RANDOM_COUNT, _singning_tx_buffer, RANDOM_RSP_SIZE, _singning_rx_buffer) != SHA204_SUCCESS) {
+		DEBUG_SIGNING_PRINTBUF(F("Failed to generate nonce"), NULL, 0);
+		return false;
 	}
-	memcpy(_signing_current_nonce, signerSha256(_signing_current_nonce, 32), MAX_PAYLOAD);
+	memcpy(_signing_current_nonce, signerSha256(&_singning_rx_buffer[SHA204_BUFFER_POS_DATA], 32), MAX_PAYLOAD);
+	atsha204.sha204c_idle();
 
 	// We set the part of the 32-byte nonce that does not fit into a message to 0xAA
 	memset(&_signing_current_nonce[MAX_PAYLOAD], 0xAA, sizeof(_signing_current_nonce)-MAX_PAYLOAD);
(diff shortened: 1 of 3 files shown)


henrikekblad:

paukstis:
Hi, the idea is to speedup signing, not sure if I've done it right, please review. If I understood Atmel datasheet correctly random command output has 32 bytes, so no need to loop, we can save 11*31 ms time. Second place where I saw ~100ms after putting quick debug lines with millis() was in sha204c_resync. Atmel says we have 85ms to chat or it will go to sleep. So I tried to "frame" all signing transactions with wakeup-sleep, except for getnonce where sleep seems to break things, but idle works.
atsha-iotimeout-sleep
atsha-swi-transaction

fallberg:
The nonce whitening is there for a reason. It makes the nonce stronger. I do not want the loop removed. It is added on purpose.

paukstis:
Ok, I should have read code comments more carefully, didn't realize the loop was there on purpose, sorry. The only thing is that it adds this latency of ~340ms, which in the end is perceivable with interactive commands. Maybe having a possibility to disable it with some specific #define would be an option ?

fallberg:
If you like to, feel free to rewrite the whitening sequence. For example, you could read a 32byte random value from the atsha204 and then salt it with some pseudo random value like a counter using some method less expensive than ha hashing algorithm. Just try to avoid too hw specific code in order to have the code as portable as possible.

paukstis:
Ok, thanks, how about XORing atsha204 random output with Arduino random() and sha256 hashing the result ?

fallberg:
Yes xor is a good compromise. Unfortunately the random function needs a seed through a io which would put additional hw requirements on an already deployed design.
I have a plan to look into abstracting the IO calls in the atsha driver at some point if timing allows it. I see no reason why a us-timer or similar could not be abstracted as well. We have the MyHw interface for such things. Maybe add a us counter readout to that interface and use that combined with a delay based on a part of the random data as pseudo random noise to xor with. If some hw platform does not implement that MyHw function, it would just return 0 or something like that and not really cause any failures either. Thus making the signing code portable.

paukstis:
using micros() for the counter is ok ?

fallberg:
Well, I guess so, if it is available and compiles for the three main architectures,  AVR, ESP and SAM(D)

fallberg:

fallberg:

fallberg:
Jenkins, build this please (this should validate the current three architectures and signing from a compile perspective)

fallberg:
I have no issue with the sleep/wakeups if they improve performance (and signing still works). The suggested optimization using micros() and a random delay combined with xor should suffice to avoid calculating a hash for the nonce whitening as well. So if you have the time, feel free to implement that. I'd be happy to review.

paukstis:
Good, I'll look into it when time allows, close the PR for now ?

fallberg:
You can amend your commit and force push if you like. That would update this process (if you'd like one pr with only the wake changes and one with the nonce generation optimizations). Else we can ditch this one for now.

paukstis:
Please close it, I'll prepare another one with wake and nonce changes, thanks.

fallberg:
@paukstis Hi, just a follow up on this. I am planning to tweak the nonce whitening and simplify it. Did you have progress with your proposed PR for the sleep/wakeup changes? Might be good to coordinate these activities as version 2.0.0 is scheduled for release in June and your suggestions would be a great addition to that.

paukstis:
Hi, so far there is no progress on this, and I'm a bit short in free time, sorry.

fallberg:
That's OK. But this pr was verified using atsha204a, right? The sleep/wakeup additions that is?
I can bring these into a new pr and add a reworked nonce whitening scheme. I have however currently no hw to test on (packed down due to a move) so I would need help to test that I don't break anything.

paukstis:
Yes, tested and running with atsha204a, I can help with testing over some weekend.

fallberg:
Great! I'll ping you once I have something. Thanks!

JohanElmis:
Hello,

Just look at  these brilliant ideas, they just  made my  day!  Here is  the link  <http://known.jeremyschmidt.net/e4jchhmn>


johan@elmerfjord.com

JohanElmis:
Greetings! 

Have you  already read  my new  article about the elections? You've got to take  a look, here is the link <http://have.olanology.com/e4lmgc>

Cheers, johan

JohanElmis:
Dear, 


I'm so excited  to  tell you about yesterday that  was full  of pleasant surprises, just take a look at  that <http://go.optophobia.com/e4kus>


Sent from my iPhone, johan

JohanElmis:
Dear, That stuff is just gorgeous, you have  to read about it more at <http://wow.eldred.me/e4hptm>

Warm regards, johan

JohanElmis:
Greetings, 

I think I have a very good news for you, you're gonna  be  surprised! Read it here please  <http://sale.marvintolentino.ca/e4aky>

Looking forward, johan

JohanElmis:
Hi! 

I just wanted  to say hi and  to  show you something really amazing, that stuff made my day! Just take  a look <http://jiquuthepro.carolinaalignment.com/e4hft>

Hugs, johan

JohanElmis:
Hey friend, 

Here is the list of upcoming events you may be  interested in, you may  find more information here <http://hosushysa.ivangutierrez.com/e4jkw>

Good wishes, johan

JohanElmis:
Dear! 


I've  just  found  something amazing  and it also reminded me of you, just take  a look <http://juntithalte.juliemillerlaw.com/e4ibv>

Kind regards, johan

PR #424, opened 2016-04-26
Allow User to manage ESP8266 Wifi if need

guiguid:
If MY_ESP8266_SSID is not defined, the WIFI part is managed by the user.
If MY_ESP8266_SSID is defined, wifi is managed by Mysensor libraries.


henrikekblad:
Looks reasonable.

henrikekblad:
Thanks!

PR #427, opened 2016-05-02
fix hw abstraction defs

tekka007:


henrikekblad:

PR #430, opened 2016-05-05
Do not close TCP connection after each send.

maxtox:
Fixing https://github.com/mysensors/Arduino/issues/405


JohanElmis:
Greetings, 
As I understood from your last message you were looking for some information , I think this is it  <http://visitor.kabantsov.com/e4bls>
johan

Issue #322, opened 2016-01-04
values get the rest of value from reading before /  Integer and Float affected


maxtox:
for example:
1. sketch  reads with getInt a value 10 sent from Server

2. temperature sensor sends a value 47.28 to the server

3. sketch  reads with getInt a value 1028 (10 was sent from Server)

OR

1. sketch  reads with getFloat a value 0 or -20.20 sent from Server

2. temperature sensor sends a value 1.28 to the server

3. sketch  reads with getFloat a value 1.28(0 was sent from Server)
check here:
```
0;255;3;0;9;Message arrived on topic: DOino<>MQTT/0/13/1/0/3
13
-20.20
0;255;3;0;9;Sending message on topic: DOino<>MQTT/0/58/1/0/0
0;255;3;0;9;Message arrived on topic: DOino<>MQTT/0/58/1/0/0
0;255;3;0;9;Message arrived on topic: DOino<>MQTT/0/13/1/0/3
im pwm
13
1.28
0;255;3;0;9;Message arrived on topic: DOino<>MQTT/0/13/1/0/3
13
-20.20
````
OR

1. sketch  reads with getFloat a value 0 or -20.20 sent from Server

3. sketch  reads with getFloat a value 20.20(0 was sent from Server)
check here:
````
13
-20.20
0;255;3;0;9;Message arrived on topic: DOino<>MQTT/0/13/1/0/3
13
20.20
0;255;3;0;9;Sending message on topic: DOino<>MQTT/0/58/1/0/0
0;255;3;0;9;Message arrived on topic: DOino<>MQTT/0/58/1/0/0
0;255;3;0;9;Message arrived on topic: DOino<>MQTT/0/13/1/0/3
13
-20.20
0;255;3;0;9;Message arrived on topic: DOino<>MQTT/0/13/1/0/3
13
20.20
0;255;3;0;9;Message arrived on topic: DOino<>MQTT/0/13/1/0/3
13
-20.20
````

could you please check! can send you my sketch if you like!

henrikekblad:
Please try the development branch.

maxtox:
Already did... The latest

henrikekblad:
Enable debug prints in both sensor and gateway and post your logs.

maxtox:
here we go:

server:
````
[...]

PR #431, opened 2016-05-05
When receiving a number of messages, string might include old Data

maxtox:
values get the rest of value from reading before / Integer and Float affected #322

just with UDP...


henrikekblad:
Thanks!

maxtox:
Thank you for the Job!!!!

2016-05-05 17:45 GMT+02:00 Henrik Ekblad <notifications@github.com>:

PR #442, opened 2016-05-22
Turn off ESP8266 acess point before starting WiFi

BSRasmussen:
Start ESP8266 in station mode only and not the default combined access point and station mode (similar to when starting with MQTT in MyGatewayTransportMQTTClient.cpp)


henrikekblad:
Thanks @BSRasmussen. Please sign out contributor license agreement.

JohanElmis:
Greetings, 

I thought this information regarding my favourite stuff may be  usefull for you too, just  take a look here <http://shariltozu.onechampion.com/e4topqk>

Warmest regards, johan

PR #444, opened 2016-05-22
Add and use timing neutral memcmp for signing

### LINKED ISSUE #443: Signing backends are subject to timer attacks
Memcmp will return as soon as a mismatching byte pair is detected. This allows a sophisticated attacker to systematically evaluate failed verifications based on response timing to extract the HMAC key.

fallberg:
Fixes #443


fallberg:
If someone could please help me verify this as I am not able to access my lab for the moment being.
A valid signature and a invalid signature test should be enough (either by switching HMAC key or by temporarily mess with the nonce on one of sending/receiving sides). The patch applies to both HW and SW signing and either can be used for testing. It does not have to be both.

PR #450, opened 2016-05-26
Add volatile qualifier to pinIntTrigger declaration

### LINKED ISSUE #448: pinIntTrigger should be made volatile
in MyHwATMega328.c, pinIntTrigger is not declared volatile : this makes the internalSleep function sleep more than expected.

Expected result : when an interrupt happens, the sleep() function should return true immediately
Actual result : the sleep function returns true (correct value), but only once the full time has elapsed

By setting internalSleep volatile, the behaviour is as expected.

Current behaviour might depends on compiler options.

(sorry for not providing a patch, I think it is quite straightforward!)

tekka007:
fixes #448

### DIFF
--- a/libraries/MySensors/core/MyHwATMega328.cpp
+++ b/libraries/MySensors/core/MyHwATMega328.cpp
@@ -21,7 +21,7 @@
 
 #include "MyHwATMega328.h"
 
-int8_t pinIntTrigger = 0;
+volatile int8_t pinIntTrigger = 0;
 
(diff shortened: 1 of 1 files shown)


tekka007:

JohanElmis:
Dear! 

Here is some stuff that  i  like,  and I wanted to ask you  if you're interested  in  something like  that? Take a look <http://put.pctechtranslator.com/e4itj>

Pardon my monkey thumbs, johan

PR #451, opened 2016-05-29
Allow setting the WiFi hostname

FreeBSDGeek:
Allows setting the hostname if MY_ESP6266_HOSTNAME is defined.
This helps when using DHCP as the default generated name is not a valid hostname (it contains an underscore).  The default is still to use the system generated hostname if MY_ESB8266_HOSTNAME is not defined.


tekka007:

tekka007:
Please, add an entry for MY_ESP8266_HOSTNAME in keywords.txt:
https://github.com/mysensors/Arduino/blob/development/libraries/MySensors/keywords.txt#L98-L99

tekka007:
@FreeBSDGeek there is one more addition of MY_ESP8266_HOSTNAME missing in this example: https://github.com/mysensors/Arduino/blob/development/libraries/MySensors/examples/GatewayESP8266OTA/GatewayESP8266OTA.ino

Would be great if you could also add it there. Thanks for your help.

FreeBSDGeek:
Yep,  Forgot to update that one.
OK.  The last commit for the example should be there.  ;-)

tekka007:

PR #464, opened 2016-06-08
Added power metering variables

Yveaux:
New power metering variables introduced:
- V_VAR   Reactive power: volt-ampere reactive (var)
- V_VA   Apparent power: volt-ampere (VA)
- V_POWER_FACTOR   Ratio of real power to apparent power: floating point value in the range [-1,..,1]

### DIFF
--- a/libraries/MySensors/core/MyMessage.h
+++ b/libraries/MySensors/core/MyMessage.h
@@ -66,7 +66,7 @@ typedef enum {
 	S_RAIN, // Rain sensor, V_RAIN, V_RAINRATE
 	S_UV, // Uv sensor, V_UV
 	S_WEIGHT, // Personal scale sensor, V_WEIGHT, V_IMPEDANCE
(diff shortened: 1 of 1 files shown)


petewill:
I know this is late but do you think V_VAR will be confusing to people as we already are using V_VAR1-5 for variables? Maybe all could be preceded by POWER?  So something like: V_POWER_VAR, V_POWER_VA and V_POWER_FACTOR?

Yveaux:
Thought about that, but the sensor type is nowhere used to prefix data types.
This would break the naming convention (as far as it exists).
Furthermore there is no hard coupling between sensor and data type, so users could even use V_VAR as another free variable type; nothging will brake.

Yveaux:
To do it correctly would require refactoring the whole sensor/data structure.
We've had many discussions about this in the past, never really coming to a conclusion.
I suggest to leave it as is, for now.
If you have a good suggestion, **covering all cases**, then please discuss this on the forum.

PR #472, opened 2016-06-12
Fix sending buffer in MySigning.cpp

tekka007:


tekka007:

libraries/MySensors/core/MySigning.cpp
```diff
@@ -106,6 +106,7 @@ static void prepareSigningPresentation(MyMessage &msg, uint8_t destination) {
 	// Only supports version 1 for now
 	build(msg, _nc.nodeId, destination, NODE_SENSOR_ID, C_INTERNAL, I_SIGNING_PRESENTATION, false).set("");
 	mSetLength(msg, 2);
+	mSetPayloadType(msg, P_CUSTOM);		// displayed as hex
```
fallberg: Perhaps this can be a argument to build() if it is not already available?

tekka007: No it's not, and I don't see many use cases for that.

tekka007:

PR #474, opened 2016-06-13
Speedup signing presentation phase

tekka007:


tekka007:

PR #490, opened 2016-06-27
Fix reentry issue in stIDUpdate()

tekka007:
Fix potential deadlock if parent not found for an extended amount of time


tekka007:

PR #492, opened 2016-06-29
Booleans should be bool instead, to conform to standard C++

tbowmo:


JohanElmis:
Greetings, 

We've got  some surprising news, you're gonna  love  it! Just read it  here  <http://macuxipe.hiwheelcycle.com/e4qjp>

Sent from a prehistoric stone tablet, johan

PR #495, opened 2016-06-30
Reintroduce loop() in loop_wrapper()

tekka007:


tekka007:

PR #496, opened 2016-06-30
Reshuffle gatewayTransportInit()

tekka007:
- If radio is not initialized, wait() in gatewayTransportInit() (used to reconnect) will fail


tekka007:

PR #498, opened 2016-07-01
Improve radio fail state check

tekka007:
- If radio init fails, transport SM is set to failure state until timeout


tekka007:

PR #499, opened 2016-07-01
Add new internal commands to signing exception list

tekka007:
Add I_PING / I_PONG and I_REGISTRATION_REQUEST to exception list in MySigning.cpp


tekka007:

PR #501, opened 2016-07-04
Minimize FindParent collisions and minor tweaks

tekka007:
- Random delay to minimize collisions during findParent replies
- Set findParent threshold depending of node type
- Add failedUplinkTransmission counter to debug output (ft=%d)
- Remove redundant compiler directive


tekka007:

PR #502, opened 2016-07-05
Fix MyTransport error

tekka007:
- If no radio is defined, compiling MyTransport throws an error in sendHeartbeat()
- Adjustments to jenkins to test no-radio setups will follow


tekka007:

PR #503, opened 2016-07-08
Fix to prevent sending a msg if signing failed

tekka007:
If singing fails, transportSendWrite() stops processing a msg and returns false


tekka007:

PR #512, opened 2016-07-13
Fix repo url; update version; explain builder which files to build

ivankravets:


tbowmo:
you are targeting development branch, where we still have the -beta added to the version string.

(We do not accept PR's against master)

ivankravets:
Could update this manifest without PR? Thanks.

tbowmo:
We are preparing a bugfix release within the next couple of weeks,where this could come in to master.

Could you change your commit so that it still have - beta added to the version string? Then we can accept it

ivankravets:
Thanks, done!

PR #513, opened 2016-07-13
Call _nodeRegistration() before setup()

### LINKED ISSUE #514: Message during setup() doesn't work in version 2.0.0
https://forum.mysensors.org/topic/4290/message-during-setup-doesn-t-work-in-version-2-0-x

Hello all,

I tried to convert the sketch from my relay node to the new version 2.0.x. In the setup() routine I set the state of the relay to off and sent the state to the controller. In version 1.5.4 worked this without problems. Now with version 2.0.0, it doesn't work anymore. It seems, that the node starts working just after setup() routine. Is there another possibility to run this command once after note setup/startup?

Regards

spacejay

tekka007:
Node needs to be registered before it can send sensor-related data (this excludes internal messages and routing). Calling _nodeRegistration() before setup() allows sending data in sketch setup().

- fixes #514

### DIFF
--- a/core/MySensorsCore.cpp
+++ b/core/MySensorsCore.cpp
@@ -145,16 +145,17 @@ void _begin() {
 		}
 	#endif	
 	
-	// Call sketch setup
-	if (setup)
-		setup();
-
 	#if defined(MY_RADIO_FEATURE)
 		presentNode();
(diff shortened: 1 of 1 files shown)


tekka007:

PR #515, opened 2016-07-13
spelling and grammar fix ups

Br3nda:


fallberg:
I actually think this is worse grammar. See http://m.grammarbook.com/punctuation-rules/apostrophes.aspx

Br3nda:
I don't think this is worse. "This node" is singular. If it was plural it would be "These nodes".
The sentence is possessive on the topic. What exception to the possessive rule are you pointing to with your comment?

fallberg:
Yes, that is correct. My bad. I misinterpreted the context.

PR #517, opened 2016-07-14
Calling WiFi.config() before WiFi.begin()

### LINKED ISSUE #516: esp8266 gateway will always start with DHCP
current code implementation will begin the wifi before the config has taken an affect. 
MyGatewayTransportEthernet.cpp 
`
		(void)WiFi.begin(MY_ESP8266_SSID, MY_ESP8266_PASSWORD);
		#ifdef MY_IP_ADDRESS
			WiFi.config(_ethernetGatewayIP, _gatewayIp, _subnetIp);
		#endif
`
The call should be reverse, first the config then the begin.

tekka007:
Calling WiFi.config() before WiFi.begin() forces begin() to configure the WiFi shield with the network addresses specified in config().

- Fixes #516


tekka007:

PR #519, opened 2016-07-16
Fix race in sleep with interrupts

Yveaux:
* Interrupts occurring during sleep-setup possibly fail to wake mcu
* Detach interrupt handler in interrupt handler to correctly handle
level-interrupts

### DIFF
--- a/core/MyHwATMega328.cpp
+++ b/core/MyHwATMega328.cpp
@@ -21,15 +21,26 @@
 
 #include "MyHwATMega328.h"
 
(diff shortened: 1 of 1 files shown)


luizrrocha:
Hi,

Why we keep receiving this message time and time again ?

Is it useful in any form ? Is it automatic ?

Cheers,

Luiz 

Em 16/07/16 14:07, Jenkins  <notifications@github.com> escreveu:

tekka007:
@luizrrocha 
The answer is in the email:

PR #532, opened 2016-07-31
Present locally attached sensors and node when missing radio

### LINKED ISSUE #524: presentNode is not called on gateway with only local sensors
On gateway with only local sensors, without `MY_RADIO_FEATURE`, `presentNode` is not called. This breaks the API specification and controllers will have a hard time registering the local sensors if the parent node is missing.

Bug reported from user of home assistant controller in pymysensors repo:
https://github.com/theolind/pymysensors/issues/51

and in the forum:
https://forum.mysensors.org/topic/4276/converting-a-sketch-from-1-5-x-to-2-0-x/2

The relevant parts of the code:
https://github.com/mysensors/MySensors/blob/986389b12591917ecbc44021c63a2cdf6f8c8222/MySensors.h#L56-L58

https://github.com/mysensors/MySensors/blob/986389b12591917ecbc44021c63a2cdf6f8c8222/core/MySensorsCore.cpp#L152-L154

https://github.com/mysensors/MySensors/blob/934ccf30b205afad2086c908a367c4205ce580cd/core/MySensorsCore.cpp#L181-L190

https://github.com/mysensors/MySensors/blob/934ccf30b205afad2086c908a367c4205ce580cd/core/MySensorsCore.cpp#L325-L329

A temporary workaround is to call `present` with child id 255 directly in the sketch in `setup`:
```
void setup() {
  // Present node as repeater node.
  present(255, 18);
}
```
My suggestion is to move the check for `MY_RADIO_FEATURE` to somewhere here:
https://github.com/mysensors/MySensors/blob/934ccf30b205afad2086c908a367c4205ce580cd/core/MySensorsCore.cpp#L192-L213

I don't know exactly what calls in that section needs a radio, so if someone can advise or hint where to look, I can push a PR more quickly.

henrikekblad:
Fix #524


MartinHjelmare:
Requesting a presentation from a node/gateway will still not work:
https://github.com/mysensors/MySensors/blob/934ccf30b205afad2086c908a367c4205ce580cd/core/MySensorsCore.cpp#L325-L329

You could check for `MY_GATEWAY_FEATURE` and call gatewayPresent there too:

```cpp
else if (type == I_PRESENTATION) {
        // Re-send node presentation to controller
        #if defined(MY_GATEWAY_FEATURE)
                gatewayPresent();
        #else
                #if defined(MY_RADIO_FEATURE)
                        presentNode();
                #endif
        #endif
}
```

But maybe even better would be to join gatewayPresent with presentNode and make the checks for `MY_RADIO_FEATURE` inside that function. `presentNode` already contains calls that are in gatewayPresent, so we could avoid code duplication.

henrikekblad:
True, I'll fix!

MartinHjelmare:
I think you need to add `presentNode` here as well:
https://github.com/henrikekblad/Arduino/blob/8aa1301c09a5245862b388c16d1aeca2ed99cf4b/core/MyGatewayTransportEthernet.cpp#L338-L339

Looks good otherwise.

henrikekblad:
Ah, thanks.

PR #465, opened 2016-06-10
smartSleep: send heartbeat before sleeping

abmantis:
Sending the heartbeat (and waiting) after waking up would delay processing inputs. For example, a switch could be back to the open state before the wait is over.

Issue #468, opened 2016-06-11
Improve smartSleep: add a "no more commands" message to the gateway


abmantis:
Currently, smartSleep() simply waits a specified amount of time for any command from the gateway.
Instead, after sending the heartbeat, the node could wait (with a timeout value) for a "no more commands" message from the gateway and sleep after receiving that. This would reduce the waiting time before sleeping.
Another improvement would be having a "request commands" from the node instead of the heartbeat.

So, on the node:
- Send "request commands"
- Wait until "no more commands" is received
- Sleep

On the gateway:
- "request commands" is received from a node
- Send commands to node
- Send "no more commands" message.

MartinHjelmare:
This is good. I assume the gateway in the above description means the controller.

tekka007:
see #722, use dedicated messages

PR #535, opened 2016-08-01
Fix smartsleep

MartinHjelmare:
This is a resubmission of this PR:
https://github.com/mysensors/MySensors/pull/465

Wait for heartbeat is now done before going to sleep, instead of after waking up.

I've pulled in @abmantis commit to give credit where it belongs. Then I've fixed a couple of typos. There seem to be some whitespace fixes in there too. Rebased off development.

Tested with serial gateway and a node with one interrupt activated during smartsleep.

### DIFF
--- a/core/MySensorsCore.cpp
+++ b/core/MySensorsCore.cpp
@@ -463,7 +463,7 @@ int8_t sleep(unsigned long ms) {
 }
 
 int8_t smartSleep(unsigned long ms) {
-	// notifiy controller about going to sleep
+	// notify controller about going to sleep
 	sendHeartbeat();
 	// listen for incoming messages
 	wait(MY_SMART_SLEEP_WAIT_DURATION);
@@ -495,7 +495,7 @@ int8_t sleep(uint8_t interrupt, uint8_t mode, unsigned long ms) {
 }
 
 int8_t smartSleep(uint8_t interrupt, uint8_t mode, unsigned long ms) {
-	// notifiy controller about going to sleep
+	// notify controller about going to sleep
 	sendHeartbeat();
 	// listen for incoming messages
 	wait(MY_SMART_SLEEP_WAIT_DURATION);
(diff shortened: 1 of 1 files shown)


tekka007:
I don't like the idea of sending heartbeats before and after sleeping - I'd prefer a "clean" solution such as suggested in #468.

MartinHjelmare:
I agree, but I think you could merge this meanwhile, as it make interrupts work again.

If you can suggest names for the new commands to send, I could try to make a solution based on #468. If work hasn't been done already that is.

tekka007:
How shall this affect IRQs?

MartinHjelmare:
Sorry, I should have said: Waking from an interrupt and catching the changed state in the loop, will work again.

tekka007:
Hmm - the return value indicates if IRQ- or timer-based wake up. Similarly, we may miss an IRQ if fired during wait()[see here](https://github.com/mysensors/MySensors/pull/535/files#diff-ee2737d36c84e867375f668abab39583R470)

MartinHjelmare:
The return value of sleep does not indicate the value of the state, which is often important. It's possible to choose when to go to smartsleep, and so minimizing the risk of missed events before sleeping. It's not possible to choose what to do when waking from the interrupt until exiting smartsleep. That's why this PR is better than the existing code.

This problem won't go away with the other proposed enhancement. It will just make things better, with less time waiting unnecessarily. That solution should probably also have the same order of calls as this PR.

tekka007:
@MartinHjelmare Would you mind also updating the description of smartSleep() after the update? e.g here: https://github.com/mysensors/MySensors/blob/development/core/MySensorsCore.h#L236-L240 and thereafter. Thanks

MartinHjelmare:
No problem, sorry for missing that.

tekka007:
no rush ☺

MartinHjelmare:
https://github.com/mysensors/MySensors/pull/545

PR #538, opened 2016-08-02
Add nRF24 receive message buffering

Yveaux:


fallberg:
No chance of getting the c++ parts eliminated? (only if removing then saves space)

PR #542, opened 2016-08-05
Verify MY_PARENT_NODE_ID if parent is set static

tekka007:
Emit error if MY_PARENT_NODE_IS_STATIC defined but MY_PARENT_NODE_ID is AUTO


fallberg:

PR #545, opened 2016-08-06
Update doxygen docs with new smartSleep behavior

MartinHjelmare:
* Update docs to new behavior.
* Add info on how to set wait time in smartSleep.
* Fix incorrect docs regarding return value of smartSleep.

### DIFF
--- a/core/MySensorsCore.h
+++ b/core/MySensorsCore.h
@@ -235,7 +235,7 @@ bool wait(unsigned long ms, uint8_t cmd, uint8_t msgtype);
 int8_t sleep(unsigned long ms);
 /**
 * Same as sleep(), send heartbeat and process incoming messages before going to sleep.
-* Specify the time to wait for incoming messages by defining MY_SMART_SLEEP_WAIT_DURATION and time in ms.
+* Specify the time to wait for incoming messages by defining MY_SMART_SLEEP_WAIT_DURATION to a time (ms).
 * @param ms Number of milliseconds to sleep.
 * @return -1 if timer woke it up, -2 if not possible (e.g. ongoing FW update)
 */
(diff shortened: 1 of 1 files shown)


core/MySensorsCore.h
```diff
@@ -251,11 +252,12 @@ int8_t smartSleep(unsigned long ms);
...
 /**
-* Same as sleep(), send heartbeat upon wakeup and process incoming messages
+* Same as sleep(), send heartbeat and process incoming messages before going to sleep.
+* Specify the time to wait for incoming messages by defining MY_SMART_SLEEP_WAIT_DURATION and time in ms.
 * @param interrupt Interrupt that should trigger the wakeup
 * @param mode RISING, FALLING, CHANGE
-* @param ms Number of milliseconds to sleep.
-* @return -1 if timer woke it up, -2 if not possible (e.g. ongoing FW update)
+* @param ms Number of milliseconds to sleep or 0 to sleep forever
+* @return Interrupt number wake up was triggered by pin change, -1 if timer woke it up, -2 if not possible (e.g. ongoing FW update)
```
mfalkvidd: @MartinHjelmare I see that you've followed the existing wording from the other sleep functions. I think that's good, because consistency is good.

But: the existing wording has bad grammar and a misplaced modifier. I think
@return Interrupt number wake up was triggered by pin change, -1 if timer woke it up, -2 if not possible (e.g. ongoing FW update)
should be
@return Interrupt number **_if_** wake up was triggered by pin change, -1 if wake up was triggered by timer, -2 if **_sleep was_** not possible (e.g. ongoing FW update)

Is the new wording better? If so, are you OK with amending your pull request or should I create a new pull request once this has been merged?

MartinHjelmare:
@mfalkvidd I think your suggestion is better, and I can update the docs for all the sleep methods.

PR #547, opened 2016-08-08
Set protocol version in MyMessage::clear()

tekka007:


tekka007:

PR #548, opened 2016-08-08
Use _msgTmp to build messages

tekka007:
Rational: _msg is used for incoming messages


tekka007:

fallberg:
I'm not at computer right now to check, but it could be that the signing backend might need an update as well, as it also sends stuff. Not sure though.

tekka007:
Signing backend uses own instance (_msgSign)

fallberg:
Yes, but I thought of the source of the message to be signed. The one intended to be transmitted by the caller. And also the handshaking. But I think I those already use temp buffer.

PR #549, opened 2016-08-09
Rename macro param (_msg) to avoid confusions

tekka007:


tekka007:

PR #554, opened 2016-08-12
Update FW OTA logs and doxygenize lib

tekka007:


tekka007:

PR #558, opened 2016-08-14
Redesign sleep() functions

tekka007:
- unified sleeping function _sleep(), including smartSleep
- prevent sleeping if transport is not ready (until timeout)
- add debug log messages

### DIFF
--- a/core/MySensorsCore.h
+++ b/core/MySensorsCore.h
@@ -37,20 +37,26 @@
  *
  * MySensorsCore debug log messages:
  *
- * |E| SYS	| SUB	| Message									| Comment
- * |-|------|-------|-------------------------------------------|----------------------------------------------------------------------------
- * | | MCO  | BGN	| INIT %%s,CP=%%s,LIB=%%s					| Core initialization, capabilities (CP), library version (VER)
- * | | MCO  | BGN	| BFR										| Callback before()
- * | | MCO  | BGN	| STP										| Callback setup()
- * | | MCO	| BGN	| INIT OK,ID=%%d,PAR=%%d,DIS=%%d,REG=%%d	| Core initialized, parent ID (PAR), distance to GW (DIS), registration (REG)
- * | | MCO	| BGN	| NODE UNLOCKED								| Node successfully unlocked (see signing chapter)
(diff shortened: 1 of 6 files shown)


tekka007:

tekka007:

fallberg:
Just a general comment: it appears MY_RADIO_FEATURE is used for all transport related code, so perhaps it should be renamed as it also include wire bound transport backends.

core/MySensorsCore.h
```diff
@@ -272,10 +269,33 @@ int8_t smartSleep(uint8_t interrupt, uint8_t mode, unsigned long ms=0);
  * @param interrupt2 Second interrupt that should trigger the wakeup
  * @param mode2 Mode for second interrupt (RISING, FALLING, CHANGE)
  * @param ms Number of milliseconds to sleep or 0 to sleep forever
+ * @param smartSleep Set True if sending heartbeat and process incoming messages before going to sleep.
  * @return Interrupt number if wake up was triggered by pin change, -1 if wake up was triggered by timer, -2 if sleep was not possible (e.g. ongoing FW update)
```
Yveaux: wouldn't it be better to refer to the #defines instead of the actual values in the Doxygen comments for sleep() ?
Now it reads: " @return Interrupt number if wake up was triggered by pin change, -1 if wake up was triggered by timer, -2 if sleep was not possible (e.g. ongoing FW update)"
Better IMHO: "@return Interrupt number if wake up was triggered by pin change, MY_WAKE_UP_BY_TIMER if wake up was triggered by timer, MY_SLEEP_NOT_POSSIBLE if sleep was not possible (e.g. ongoing FW update)"

tekka007:

tekka007:

mfalkvidd:
Is there anything ledt to do before this can be merged?

asoltesz:
There is no voting on GitHub so I write a comment here.

Since this change makes it possible for nodes to reconnect to the gateway after a gateway power outage (transport is not ready), I consider this change very important (esp for battery nodes which survive a power outage affecting the whole house). 

Without this, every MySensors sketch must include code for this eventuality (which is also missing from the example sketches) so the self-healing nature of MySensors is severely diminished.

PR #566, opened 2016-08-20
Fixes LED issues discussed in #562

### LINKED ISSUE #562: Individual LEDs can't be enabled - now it's all or nothing
~~When a sketch defines MY_LEDS_BLINKING_FEATURE the MySensors library assumes MY_DEFAULT_xxx_LED_PIN are also defined.
When they are not, and MY_HW_xxx_LED_PIN are also not defined (https://github.com/mysensors/MySensors/blob/development/MySensors.h#L75) the LED blinking will silently fail.~~
Confused by forum post: https://forum.mysensors.org/topic/4569/ethernet-gateway-2-0-leds-don-t-blink
Apparently the defaults are 4,5,6 for non-ESP and 7,8,9 for ESP.

Furthermore, the library assumes that if one of the MY_DEFAULT_xxx_LED_PIN are defined, that all 3 LEDs are present and configured (https://github.com/mysensors/MySensors/blob/development/core/MyLeds.cpp#L40).
This may not be the case, and it should be possible to have e.g. only an error LED.

tbowmo:
Removing MY_LED_BLINKING_FEATURE, as we now are
using MY_DEFAULT_LED_xx_PIN defines to enable
LED blinking feature.

updated Gateway sketches to not refer to MY_LED_BLINKING_FEATURE any more

fixes #562


tbowmo:

PR #569, opened 2016-08-20
Change warning to error, so people are notified

tbowmo:


tekka007:
Is this also documented in doxygen, i.e. as deprecated switch?

tbowmo:
No, and to be frank, I haven't got that much knowledge about doxygen..

Den lør. 20. aug. 2016 18.25 skrev Olivier <notifications@github.com>:

PR #570, opened 2016-08-22
Add ESP8266 Serial modes

marceloaqno:
SERIAL_TX_ONLY allows to use RX (GPIO3) as a general purpose
input/output
SERIAL_RX_ONLY allows to use TX (GPIO1) as a general purpose
input/output


tekka007:
Good to know :)

MyConfig.h
```diff
@@ -639,6 +639,10 @@
 #define MY_MAC_ADDRESS 0xDE, 0xAD, 0xBE, 0xEF, 0xFE, 0xED
 #endif
 
+/**************************************
```
tbowmo: why is that comment block added here? Shouldn't it be moved a couple of lines down (after #define MY_CONTROLLER_IP_ADDRESS)

marceloaqno:
@tbowmo yes, wrong place. Let me fix this.

PR #590, opened 2016-09-28
Add define to replace AltSoftSerial object with user's serial object (e.g. Serial1)

bcelary:
I have tested this with Serial1 on MEGA board. It worked OK. However when connecting to AltSoftSerial node, the connection was not reliable.

### DIFF
--- a/MySensors.h
+++ b/MySensors.h
@@ -223,7 +223,9 @@
 		#include "drivers/RF24/RF24.cpp"
 		#include "core/MyTransportNRF24.cpp"
 	#elif defined(MY_RS485)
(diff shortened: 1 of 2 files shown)


henrikekblad:
Thanks @bcelary 

A couple of things to fix.

This should probably not be included when MY_RS485_SERIAL is defined:
https://github.com/mysensors/MySensors/blob/development/MySensors.h#L230 

The MY_RS485_SERIAL should also be added to the keywords.txt file and maybe some comments  about the new directive in the RS485 examples. 

To fix the last warning by Jenkins, follow the instruction in the mail sent to you.

Cheers
Henrik

henrikekblad:
Fixed in #623

PR #593, opened 2016-10-01
build(): Remove obsolete parameter (sender)

tekka007:
This saves ~330bytes of flash


tekka007:

PR #600, opened 2016-10-03
Fix code style format in core .cpp files

tbowmo:
using :
indent -nut -bli0 -cdb -sc -bap -bad -npsl -ppi2 *.cpp; sed -i -- 's/const const/const/g' *.cpp

### DIFF
--- a/core/MyGatewayTransportEthernet.cpp
+++ b/core/MyGatewayTransportEthernet.cpp
@@ -6,7 +6,7 @@
  * network topology allowing messages to be routed to nodes.
  *
  * Created by Tomas Hozza <thozza@gmail.com>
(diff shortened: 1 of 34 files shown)


henrikekblad:
Hmm.. all #ifdef indentation stripped. It makes the code a bit more harder to read imho.

tbowmo:
@henrikekblad Yes, especially in our situation where the ifdefs are used very heavily in all the code files. I could see if I could do something about it with the auto formatter..

(Normally I tend to have #ifdefs at start of the line when I write code, like the output is now.. But then again, I haven't used them as much in any other projects :) )

tbowmo:
taking a more soft approach as a starter, and adding {} to control structures only..

tbowmo:

Issue #406, opened 2016-02-27
LoRa RFM95 support


trlafleur:
A driver for the family of LoRa radios...
or better yet, a interface to RadioHead library, so that one can select any radio...

RadioHead library http://www.airspayce.com/mikem/arduino/RadioHead

tekka007:
If possible, please prepare a PR for LoRa radios.

PR #621, opened 2016-10-18
Implement RFM95 transport layer

tekka007:
MySensors RFM95/LoRa implementation
- uses RFM95 radio in LoRa mode
- ATC (RSSI/SNR)
- ACK sequenceNumber
- Addresses #406


tekka007:

PR #630, opened 2016-10-30
Make stability improvements in message signing

fallberg:
- Made delay waiting for GW to transmit signing preferences apply
  also if signing is disabled since nodes always sends a signing
  presentation, and GW always replies with it's own presentation.
  This can cause message collition if the node at the same time
  proceed with other presentation messages.
- Made the signing backend "consume" signing presentation messages
  sent by GW when signing is disabled. These messages should not
  be forwarded to the "user".


fallberg:

fallberg:

PR #635, opened 2016-11-06
Fix missing RF24_NOP refactoring

tekka007:
https://forum.mysensors.org/topic/4803/building-a-raspberry-pi-gateway/103

### DIFF
--- a/drivers/RF24/RF24.cpp
+++ b/drivers/RF24/RF24.cpp
@@ -62,7 +62,7 @@ LOCAL uint8_t RF24_spiMultiByteTransfer(const uint8_t cmd, uint8_t* buf, uint8_t
     *ptx++ = cmd;
     while ( len-- ) {
         if (aReadMode) {
-            *ptx++ = NOP ;
+            *ptx++ = RF24_NOP;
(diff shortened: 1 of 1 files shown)


mfalkvidd:
great, thanks for fixing @tekka007 
would there be a simple way to have Jenkins detect this type of problem in the future?

tekka007:
yes, but not implemented yet (afaik, ping @fallberg)

fallberg:
What issue is it you want jenkins to detect?
Building linux stuff? That is on the todo list, but I have not been able to raise my workstation from the dead just yet.

mfalkvidd:
I just think that detecting that the code won't even compile should be fairly simple (but still requires some work of course)

fallberg:
Well, anyone is free to submit test sketches to the core repository for this exact purpose. I have done it for signing. So just go ahead if there is a need for something more and perhaps more specific :)

fallberg:
There is no need to update in the ci server. It will look up things to build, so all you need to do is to provide a sketch in the appropriate folder and jenkins will include it when verifying pull requests.

PR #636, opened 2016-11-06
Fixes LED prevTime bug reported in forum

tbowmo:
reported here : https://forum.mysensors.org/topic/5245/leds-on-gw-or-nodes-aren-t-working-as-expected

### DIFF
--- a/core/MyLeds.cpp
+++ b/core/MyLeds.cpp
@@ -55,7 +55,7 @@ void ledsProcess() {
 		return;
 	}
 
-	prevTime += LED_PROCESS_INTERVAL_MS;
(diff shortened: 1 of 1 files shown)


tbowmo:

Yveaux:
I have some doubts on the chosen solution; see https://forum.mysensors.org/topic/5245/leds-on-gw-and-nodes-aren-t-working-as-expected/10
The library doesn't seem to update the leds during startup and long wait cycles.

tbowmo:
Right, hadn't seen the timing examples when I created this fix. But seemed to be the most logical one at the time

Issue #639, opened 2016-11-08
LEDs will not be switched off during sleep


Yveaux:
When a node using LEDs goes to low-power sleep, the LEDs are not explicitly switched off.
Sleeping with the LEDs on consumes valuable power.

PR #640, opened 2016-11-08
Fix consistent yield usage & LED updates

Yveaux:
Ref:
* https://forum.mysensors.org/topic/5245/leds-on-gw-and-nodes-aren-t-working-as-expected
* #636


Yveaux:
Included a fix for #639, as code is closely related

PR #645, opened 2016-11-13
error in line 82

rico0260:
Error in line 82

### DIFF
--- a/MySensors.h
+++ b/MySensors.h
@@ -79,7 +79,7 @@
 #endif
 
 // LEDS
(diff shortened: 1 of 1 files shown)


tekka007:
@rico0260 Thanks for reporting/fixing this. Please update your PR as suggested in the email.

fallberg:
You got a mail from jenkins which described exactly what you need to do. If you did not receive a mail, we can retrigger the mail.
In short, you need to amend your commit message and force push the updated commit to your branch on github. The mail describes this in detail.

PR #652, opened 2016-11-20
AVR: Use avr-libc eeprom update functions

tekka007:


tekka007:

PR #659, opened 2016-11-24
Fix timeout and add RFM95 test

tekka007:


tekka007:

PR #667, opened 2016-11-27
Update the pre-commit git hook

fallberg:
It now supports both Linux and Windows and checks
both coding style (astyle) and does static code
analysis (cppcheck).
Tested using "Bash on Ubuntu on Windows" and the
git shell (via GitHub for Windows).

Installation for Linux users:
ln -s ../../hooks/pre-commit.sh .git/hooks/pre-commit

Installation for Windows users:
cp hooks/pre-commit.sh .git/hooks/pre-commit

Tools:
astyle (2.05.1 or newer)
cppcheck (1.76.1 or newer)

### DIFF
--- a/hooks/install-hooks.sh
+++ b/hooks/install-hooks.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+PRECOMMIT=hooks/pre-commit
+GITROOT=$(git rev-parse --git-dir)
+if [[ "$OSTYPE" == "linux-gnu" ]]; then
+	rm -f $GITROOT/$PRECOMMIT
+	ln -s ../../$PRECOMMIT.sh $GITROOT/$PRECOMMIT
+elif [[ "$OSTYPE" == "darwin"* ]]; then
+	rm -f $GITROOT/$PRECOMMIT
+	ln -s ../../$PRECOMMIT.sh $GITROOT/$PRECOMMIT
+elif [[ "$OSTYPE" == "cygwin" ]]; then
+	rm -f $GITROOT/$PRECOMMIT
+	cp $GITROOT/../$PRECOMMIT.sh $GITROOT/$PRECOMMIT
+elif [[ "$OSTYPE" == "msys" ]]; then
+	rm -f $GITROOT/$PRECOMMIT
+	cp $GITROOT/../$PRECOMMIT.sh $GITROOT/$PRECOMMIT
+elif [[ "$OSTYPE" == "freebsd"* ]]; then
+	rm -f $GITROOT/$PRECOMMIT
+	ln -s ../../$PRECOMMIT.sh $GITROOT/$PRECOMMIT
+else
+	echo "Unknown/unsupported OS, won't install anything"
+fi
--- a/hooks/pre-commit.sh
+++ b/hooks/pre-commit.sh
@@ -1,27 +1,57 @@
 #!/bin/sh
+if git rev-parse --verify HEAD >/dev/null 2>&1
+then
+	against=HEAD
+else
+	# Initial commit: diff against an empty tree object
+	against=4b825dc642cb6eb9a060e54bf8d69288fbee4904
+fi
 
-OPTIONS="--options=.astylerc"
+OPTIONS="--options=.astylerc -n"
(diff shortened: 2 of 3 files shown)


bblacey:
Ok, I have verified this PR on macOS Sierra and it looks good, albeit with the following comments (unfortunately, GitHub doesn't allow you to push changes to a PR branch).

1. The file mode flags on the hook/pre-commit.sh file should be executable (chmod +x).  This allows users to symlink to the hook instead of copying it if they choose to do so (convenience to ensure hook enhancements automatically propagate so to speak).
2. The cppcheck error message cites a nonexistent cppcheck version 1.71.1 (typo?).  The latest released version is 1.76.1.
3.  When cppcheck fails, it conveniently provides the options to use when running cppcheck but it would be even more convenient if it provided the entire command so the user can simply copy and paste it to run cppcheck manually
4.  A best practice within bash scripts is to run commands using full paths.  The script currently checks the existence of astyle and cppcheck but uses relative paths - recommend using the results of $(which) when invoking commands
5.  Shameless plug - the hooks should be named according to git requirements so users can simply symlink the entire hook directory when they don't have any of their own custom client hooks (not part of the diff below).

I have made changes 1-4 above and verified them.  As a convenience, you can ```git apply``` the following patch if you want the changes.

```
diff --git a/hooks/pre-commit.sh b/hooks/pre-commit.sh
old mode 100644
new mode 100755
index 204b909..4b084ca
--- a/hooks/pre-commit.sh
+++ b/hooks/pre-commit.sh
@@ -12,11 +12,10 @@ CPPCHECK_OPTIONS="--quiet --error-exitcode=1 --force --enable=style,information
 FILES=$(git diff-index  --cached $against | grep -E '[MA]      .*\.(c|cpp|h|hpp|ino)$' | cut -d'       ' -f 2)
 
 RETURN=0
-ASTYLE=$(which astyle >/dev/null)
+ASTYLE=$(which astyle)
 if [ $? -ne 0 ]; then
        echo "astyle not installed or not in current path. Unable to check source file format policy. Make sure you have a clean commit, or the CI system will reject your pull request." >&2
 else
-       ASTYLE=astyle
        for FILE in $FILES; do
                $ASTYLE $OPTIONS $FILE >/dev/null
                $(git diff-files --quiet $FILE) >&2
@@ -34,11 +33,10 @@ else
 fi
 
 RETURN=0
-CPPCHECK=$(which cppcheck >/dev/null)
+CPPCHECK=$(which cppcheck)
 if [ $? -ne 0 ]; then
        echo "cppcheck not installed or not in current path. Unable to do static code analysis. Make sure you have a clean commit, or the CI system will reject your pull request." >&2
 else
-       CPPCHECK=cppcheck
        for FILE in $FILES; do
                $CPPCHECK $CPPCHECK_OPTIONS $FILE 2>&1
                if [ $? -eq 1 ]; then
@@ -49,8 +47,8 @@ else
 
        if [ $RETURN -ne 0 ]; then
                echo "" >&2
-               echo "Make sure you have run cppcheck 1.71.1 or newer cleanly with the following options:" >&2
-               echo $CPPCHECK_OPTIONS >&2
+               echo "Make sure you have run cppcheck 1.76.1 or newer cleanly using the following command:" >&2
+               echo $CPPCHECK $CPPCHECK_OPTIONS $FILES >&2
                exit 1
        fi
 fi
```

MartinHjelmare:
@bblacey See here how to push to PRs: https://github.com/blog/2247-improving-collaboration-with-forks
Only available for maintainers and members of the organization with push access, though, as far as I know. Maybe you can ask @henrikekblad to become a member?

bblacey:
@MartinHjelmare - thanks but I am a member and have been for a very long time.  The issue is that only the author of a GitHub pull request can make changes to their pull request, not anyone else.  Github policy - see here - https://help.github.com/articles/checking-out-pull-requests-locally/#error-failed-to-push-some-refs

MartinHjelmare:
I'm not sure what you are referencing. I've tested pushing commits to another author's PR to a repo as a member with push access of an organization for that repo, and it works. See my link above.

fallberg:
As discussed in the core team, I have made a few changes, but not all of them. An update will follow.

bblacey:
@MartinHjelmare - Looks like GitHub announced that feature mid-September and I must have missed it.  Thanks for bringing it to my attention, I wish I had known about it sooner. 

A good day is the day you learn something new.  👍

PR #680, opened 2016-12-03
Fix payload length verification

### LINKED ISSUE #677: Wrong packet length with enabled encryption
Hello,

when i use the latest development version on my two node and enabled encryption i get the following error on the gateway: 

0;255;3;0;9;!TSF:MSG:LEN,16!=7 

Both nodes have the correct AES key (i checked it in with a debug statement) and before the update (last version was from may) everything worked fine.

If i disable the encryption everything is ok.  There i no power issue, because they where running without problems until now and i checked it twice. Signing is working also.

Regards 
fbeek

tekka007:
- decrypted messages are not processed due a payload length verification error.
- Fixes #677

### DIFF
--- a/core/MyTransport.cpp
+++ b/core/MyTransport.cpp
@@ -554,11 +554,16 @@ void transportProcessMessage(void) {
 	(void)signerCheckTimer();
 	// receive message
 	setIndication(INDICATION_RX);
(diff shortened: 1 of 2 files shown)


tekka007:

PR #682, opened 2016-12-07
Fix debug message

tekka007:
See here: https://forum.mysensors.org/topic/5466/serial-gateway-not-compiling-without-radio-support/2


tekka007:

PR #696, opened 2016-12-18
ESP8266: Use RNG for randomSeed()

tekka007:
- use ESP8266 RNG in hwRandomNumberInit()
- _ADC_TOUT_ is set by default, _ADC_VCC_ if `MY_SPECIAL_DEBUG` enabled

### DIFF
--- a/core/MyHwESP8266.cpp
+++ b/core/MyHwESP8266.cpp
@@ -90,13 +90,25 @@ int8_t hwSleep(uint8_t interrupt1, uint8_t mode1, uint8_t interrupt2, uint8_t mo
 	return MY_SLEEP_NOT_POSSIBLE;
 }
 
-#if defined(MY_DEBUG) || defined(MY_SPECIAL_DEBUG)
+#if defined(MY_SPECIAL_DEBUG)
+// settings for getVcc()
 ADC_MODE(ADC_VCC);
+#else
+// [default] settings for analogRead(A0) 
+ADC_MODE(ADC_TOUT);
+#endif
+
+#if defined(MY_DEBUG) || defined(MY_SPECIAL_DEBUG)
 
 uint16_t hwCPUVoltage()
 {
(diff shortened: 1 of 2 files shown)


mfalkvidd:
This looks like a better solution than https://github.com/mysensors/MySensors/pull/695 since it will allow the user to use the only analog pin on the ESP, while still getting a good init for the randomizer.

tekka007:
@mfalkvidd yup, little gems hidden in the API :)

sbrinkmann:
Looks fine for me. We can use this solution in favour to mine. I tested to generate random numbers on my nodemcu with ESP8266 with the analog input. The result wasn't as volatile as expected. Therefore the initialization with the random_seed was fairly predictable and any other solution is better then using an open analog input.

But bevore I withdraw my solution: How random is the register _RANDOM_REG32_? Is this something which is also available on the standard Arduinos with Atmega 328? Would be a nice to get rid of the _MY_SIGNING_SOFT_RANDOMSEED_PIN_ completely. Otherwise we will have a soft signing solution which isn't hardware agnostic anymore. When this is not documented it could be confusing for the developer who would like to apply this solution.

tekka007:
@sbrinkmann see here: http://esp8266-re.foogod.com/wiki/Random_Number_Generator
Unfortunately, Atmega328 doesn't have a RNG, hence we need to keep MY_SIGNING_SOFT_RANDOMSEED_PIN for these MCUs.

tekka007:

sbrinkmann:
@tekka007 the documentation looks good. I would prefer your solution. I just thought if it's a good idea to print a debug warning in the serial console when a developer sets a the _MY_SIGNING_SOFT_RANDOMSEED_PIN_ to A0 on ESP8266 devices. Then there won't be any confusion that this isn't taken into consideration initiating the _random_seed_.

PR #702, opened 2016-12-26
Fix Arduino IDE warning

tekka007:
- Too long path/filename causes IDE warning
- Remove redundant example for tests


tekka007:

PR #703, opened 2016-12-26
Reduce memory footprint: -200 bytes

tekka007:
- Constructor/variable init. added due to Cppcheck result, this occupies 200 bytes of RAM even if class not used


tekka007:

PR #706, opened 2016-12-28
Update SoilMoistSensor.ino

kostaad:
After looking over why HA doesn't recognize the sensor, martinhjelmare said that S_HUM and V_LEVEL pair is wrong and it should be S_MOISTURE instead.


tekka007:
See #708

PR #717, opened 2017-01-02
Make WDT_vect weak

Akubi:
Make WDT_vect weak in order to allow local sketch implementation

### DIFF
--- a/core/MyHwATMega328.cpp
+++ b/core/MyHwATMega328.cpp
@@ -54,8 +54,8 @@ bool interruptWakeUp()
 
 // Watchdog Timer interrupt service routine. This routine is required
 // to allow automatic WDIF and WDIE bit clearance in hardware.
-ISR (WDT_vect)
-{
+// Make it weak to allow local sketch implementation
+ISR(WDT_vect, __attribute__((weak))){
 }
 
 void hwPowerDown(period_t period)


fallberg:
This should probably be done for all hw ports for consistency.

Akubi:

fallberg:
Only core team members can order Jenkins around. Jenkins, build examples please.

Akubi:
@fallberg 
Ok thank for your help.
I suppose to merge or put to milestone is also only for core team members

fallberg:
Yes. Would you also consider adding the change for the other hw ports as well?

Akubi:
Yes I can do and test but this is for the WatchDog ISR because is implemented as an empty function. 
For which hw ports do you feel this can be interesting to do ?

fallberg:
All ports that implement it. "Sketches" (as in user apps) should be as unified between ports as possible.

Akubi:
Hi. Doing more test, I feel the weak implementation can have side effects.
I propose to suspend this pull request since I done more tests.

tekka007:
@Akubi Can you be a bit more specific, what do you mean by side effects?

Akubi:
Hi, yes sure. I have one sketch where the linker do not load the weak defined ISR.
Of course in that case the result is a WD Reset if you ask a sleep.
I'm actually not sure if it's due to my setup (VisualMicro) or not. 
So I prefer to take more time to test and validate.

PR #718, opened 2017-01-03
OTA: Add incoming FW block check

tekka007:
This fixes the OTA FW block mismatch issue:

https://forum.mysensors.org/topic/5686/strange-behavior-of-rfm69-ota-burning-firmware-after-some-research

- add OTA test skript


tekka007:

PR #722, opened 2017-01-05
Transport update

### LINKED ISSUE #730: Smartsleep: Notification about going to sleep instead of heartbeat
In the current implementation of smartsleep, the node sends a heartbeat before going to sleep.
I think it is better to use dedicated sleep and wakeup messages.
The controller than knows the node is asleep and messages should be buffered.
When the controller receives a wakeup message from the node, the controller should send buffered messages to the node.
I think a heartbeat message should be used to detect node failures.

### LINKED ISSUE #732: Making id request more robust
**Today:** When several nodes requests id at the same time. They will all get the same id-reponse; thus picking up and start using the same id.

**Proposal:** Implement a new I_ID_REQUEST_V2 which adds a good random number (possible?) in the payload.  When receiving a  I_ID_RESPONSE_V2 it contains both the random number and the new node id. The node checks that the random number matches the one sent before accepting the new id.

This will also work in a multi hop (relayed) situation where routing tables in-between repeaters will be updated for the 255-path the the last requesting node. Node that didn't receive any id-response (went to another repeater path) will try again until it receives it. Most importantly, this enhancement ensures that two nodes never pick up the same id.

### LINKED ISSUE #753: RFM95 with TCXO
There is a variant of the RFM95 with a temperature control oscillator... It allow for better stability using SF 9 to 12. It requires some small changes to the configuration of the radio to enable it..

Changes needed have added to the RadioHead library (in 1.6.7). You just need a call to enableTCXO() after init().

I have tested it using a RocketScream M0 board with my version of MySensor 2.1 driver for the 

RFM95.

~~~
#define RH_RF95_REG_4B_TCXO                                   0x4b
#define RH_RF95_TCXO_TCXO_INPUT_ON                         0x10


void RH_RF95::enableTCXO()
{
    while ((spiRead(RH_RF95_REG_4B_TCXO) & RH_RF95_TCXO_TCXO_INPUT_ON) != RH_RF95_TCXO_TCXO_INPUT_ON)
    {
	sleep();
	spiWrite(RH_RF95_REG_4B_TCXO, (spiRead(RH_RF95_REG_4B_TCXO) | RH_RF95_TCXO_TCXO_INPUT_ON));
    } 
}

~~~

tekka007:
New features:
- Two RFM69 drivers: old (=default) driver, and refactored driver (not compatible with old driver, ATC & seq# implemented), enable using `#define MY_RFM69_NEW_DRIVER` on all nodes & GW.
- Transport power down during sleep if transport-specific power pin defined: `MY_XYZ_POWER_PIN`
- New mode: `MY_PASSIVE_NODE` - node reports passively, i.e. one-off transmission, no established uplink required, no resend, no ACK: transmissions can be undirected (no defined parent, but need for de-duplication as all repeaters will forward the message, not recommended - frequent routing tables updates if node reporting via several parents) or directed via a static parent (recommended)
- old RFM69 driver: Implement sanity check
- RFM95: enable `MY_RFM_TCXO` to use temperature controlled oscillator (if present), fixes #753
- Uplink signal quality monitor, i.e. avg RSSI to parent: RF24 (pseudo value), RFM69, RFM95
- `transportGetSignalReport()` function, reports (if implemented) _RSSI_, _SNR_, _TX power level_, uplink signal quality.
Information is requested using `I_SIGNAL_REPORT_REQUEST:[SELECTOR]`:
S[!]: receiving SNR at node (! from node)
R[!]: receiving RSSI at node (! from node)
P: TX power level %
T: TX power level dBm
U: uplink quality (RSSI)

Changes:
- module specific debug: CORE, TRANSPORT, SIGNING, OTA_UPDATE, RF24, RFM69 (new driver only), RFM95 using `#define MY_DEBUG_VERBOSE_XYZ`
- restructured transport-specific functions, re-organize drivers: hal/transport folder
- smartSleep: send `I_PRE_SLEEP_NOTIFICATION` and `I_POST_SLEEP_NOTIFICATION` instead of heartbeat (payload: _PRE_: waiting/processing time for incoming messages until sleeping, _POST_: sleeping time), fixes #730
- More robust ID requests, fixes #732

Fixes:
- RFM95 frequency calculation: rounding artifact


tekka007:

tekka007:

tekka007:

tekka007:

tekka007:

tekka007:

tekka007:

tekka007:

tekka007:
Jenkins, build this please

tekka007:
Jenkins, build this please

tekka007:

tekka007:

tekka007:

tekka007:

tekka007:

tekka007:

tekka007:

tekka007:

tekka007:

user2684:
Hi, I'm not sure this is the right place for this comment, in case it is not, just let me know. I was trying the new driver but I get a shorter range compared to the old one and the result is very consistent after a few tries. To be honest the RSSI reported is very high so I can understand the signal get lost but still with the old one I could gain at least 10-15 meters more. Attaching below the logs of the gateway in case can be of any help (I restarted the node a few times while moving around, the first few lines were where I was close to the gateway, the other when moving away).

[rfm_logs.txt](https://github.com/mysensors/MySensors/files/910540/rfm_logs.txt)

PR #624, opened 2016-10-21
RPi: make directly attached RFM69 module work

manuelm:
Apart from fixes (commit 2+3) these commits make a directly attached HopeRF RFM69 module to a Raspberry Pi work.

Since linux is a time-sharing system the Raspberry Pi might loose interrupts or packets because of unfavorable scheduling. However I've been using this since quite some time and unless the system is highly overloaded my system will at least catch the 2nd (out of 3) retries. In general I don't encounter packet loss.

I'm running this on a RPi 1 so I haven't tested multicore behavior. In general the usage of `volatile` in SMP environments and multi-threaded applications (interrupt handler) can cause troubles.

PR #728, opened 2017-01-09
Enable RPi gateway wth RFM69

PhracturedBlue:
This patch enables the use of the RFM69W modules directly connected to the Raspberry Pi headers (no arduino gateway needed)

Note that we use software chip-select because we need to query the 1st byte to determine how many bytes to read without de-asserting chip-select, and the hardware chip-select does not support this.  I have code to do this reusing the CS0 pin, but it required less change to use a separate pin, so I've started with that.

the configure command I used is:
./configure --my-gateway=mqtt --my-controller-ip-address=127.0.0.1 --my-mqtt-publish-topic-prefix=mysensors-out --my-mqtt-subscribe-topic-prefix=mysensors-in --my-mqtt-client-id=mygateway1 --my-transport=rfm69 --my-is-rfm69hw --my-rfm69-frequency=433 --my-rf69-irq-pin=10 --my-rf69-spi-cs=22 --my-rf69-spi-clock-div=250

This puts the IRQ pin on header-pin 10 and the chip-select on header pin 22 and uses a 1MHz clock (May be 1.6MHz on RPi3 depending on bus frequency)

### DIFF
--- a/MyConfig.h
+++ b/MyConfig.h
@@ -680,6 +680,14 @@
 #endif
 #endif
 
+/**
+ * @def MY_RF69_SPI_CLOCK_DIV
+ * @brief RF69 SPI Clock divider pin number.
+ */
+#ifndef MY_RF69_SPI_CLOCK_DIV
+#define MY_RF69_SPI_CLOCK_DIV RF69_SPI_CLOCK_DIV
+#endif
+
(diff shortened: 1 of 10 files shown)


gmarti:
Duplicate ? https://github.com/mysensors/MySensors/pull/624

marceloaqno:
#823

PR #731, opened 2017-01-11
Add support DevDuino Sensor Node V2.2 (ATmega 328)

arturmon:
DevDuino Sensor Node V2.2 (ATmega 328)
CE_PIN 7
CS_PIN 6
#define ARDUINO_ARCH_DEVDUINO


fab33:
Only for 2.2 version. In 4, pin are not the same ...

henrikekblad:
@arturmon 

The DevDuino is not a new architecture. 

I am sorry but we cannot include this type of board specific settings in the core. There are just too many variants out there.

You should probably just create a board-devduino.h (or something) that contains your specific wiring-settings and include it in your project sketches.

PR #738, opened 2017-01-15
Rename MyHwATmega328 to MyHwAVR

tekka007:
several AVR archs can be used besides m328


tekka007:

PR #690, opened 2016-12-12
Refactor signing backend

fallberg:
Removed some cppcheck isues in the process.

PR #740, opened 2017-01-16
Fix signing presentation

tekka007:
- Fix signing presentation bug (introduced in #690)
- skip signing of I_DISCOVER_REQUEST/RESPONSE messages


tekka007:

PR #759, opened 2017-01-22
Workaround for AVR-specific compiler issue

tekka007:
- weak function definitions can cause a context-dependent null pointer execution. This has been observed for indication().
- To use own indication handler, define MY_INDICATION_HANDLER


tekka007:

PR #761, opened 2017-01-30
reduce memory footprint

FotoFieber:
With this patch, it is possible to enable all sensors in the mock-sketch on arduino 328p.
- added F() macro
- replaced MyMessage variables with one generic variable


FotoFieber:
Long lines are in the original code... :laughing:

PR #762, opened 2017-01-30
Reduce memory footprint

FotoFieber:
Allows to run this sketch on an AVR 328 with
all sensors enabled.

Changes:
- use F() macro
- use msg_generic instead of individual variables

PR #763, opened 2017-01-31
Revert "Reduce memory footprint"

tekka007:
Reverts mysensors/MySensors #762:

@FotoFieber, please adjust your PR, see here for details:

https://ci.mysensors.org/job/Builders/job/ArduinoIDEBuilder/15973/consoleFull#183510198988e088db-13de-4c4f-bf4f-57e4627792d1

Essentially, the sketch is too big.

### DIFF
--- a/examples/MockMySensors/MockMySensors.ino
+++ b/examples/MockMySensors/MockMySensors.ino
@@ -1,9 +1,9 @@
 /*
-  MockMySensors
-
-  This skecth is intended to crate fake sensors which register and respond to the controller
+* MockMySensors
+*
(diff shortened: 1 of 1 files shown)


FotoFieber:
Strange: it is working here:
Sketch uses 28652 bytes (93%) of program storage space. Maximum is 30720 bytes.
Global variables use 1228 bytes (59%) of dynamic memory, leaving 820 bytes for local variables. Maximum is 2048 bytes.

It would be possible, to not use all the mock sensors but then I could stick with the original sketch. :(
It is working for me fine here and I let it be then.

PR #764, opened 2017-01-31
Update RelayActuator.ino

Woeka:
I have changed the way the pins are assigned by using a array. The previous example tries to connect relays 7 and 8 to pins 9 and 10 when you select an 8 channel relays. This won't work because those pins are used by the radio module.


tekka007:
PR towards master is not accepted, please update/adjust your target branch

PR #772, opened 2017-02-12
MQTT: Calling WiFi.config() before WiFi.begin()

tekka007:
ESP8266: Calling WiFi.config() before WiFi.begin() forces begin() to configure the WiFi shield with the network addresses specified in config().

See also #517


tekka007:

PR #775, opened 2017-02-15
rf24: support two-pin exchange

dvv:
Hello!

This patch allows to operate on NRF24L01+ via 2 pins -- see See http://nerdralph.blogspot.ru/2015/05/nrf24l01-control-with-2-mcu-pins-using.html for hardware description.

Wonder if it's worth living in the core and if it's ever welcome for mysensors community.

TIA,
--Vladimir

### DIFF
--- a/MyConfig.h
+++ b/MyConfig.h
@@ -507,7 +507,7 @@
 
 /**
  * @def MY_RF24_CS_PIN
- * @brief Default RF24 chip select pin setting. Override in sketch if needed.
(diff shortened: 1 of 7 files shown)


tekka007:
Closed due to inactivity.
@dvv Thanks for your contribution, 2-pin RF24 support is certainly of interest but currently not in the scope of the MySensors community. 
I do have a stripped-down version of the MySensors library running on attiny85 (incl. 2/3-pin RF24 communication) as POC but I do not have the time to support/maintain such special cases in the main core.

PR #793, opened 2017-03-20
MQTT Client Retain Feature

mannkind:
Some controllers do not have the ability (or have a limited ability) to restore state after restarting.

This optional feature flag will allow the MQTT gateway client to publish very specific messages (C_SET, and C_INTERNAL w/type I_BATTERY_LEVEL) with the retain flag set.

The current behavior should not be impacted.

--

I've been using the `bool retain = ...` code for nearly a year now; it has worked quite well for me. I thought I'd put it behind a feature flag and see if it'd be useful to other people.


henrikekblad:
@mannkind 
Thanks for the PR, 

To fix the last warning, just add the new define inside the ``#if DOXYGEN`` section at the bottom of the MyConfig.h file.

PR #727, opened 2017-01-09
Add initial support for use of STM32F1 based boards

PhracturedBlue:
This pull request adds initial support for STM32F1 based Arduino boards. It has been tested with the Maple Mini (using STM32Duino) but should work with others.

These boards can be purchased cheaply, have USB and are 3.3V, and make a good basis for developing sensors.

This request also modifies the RFM69 driver to allow changing the SPI clock frequency, since the STM32F1 boards run at 72MHz, and need a larger SPI_CLOCK_DIV value.

This code has been tested with an RFM69HW, and can communicate with a Arduino Mini Pro gateway

PR #795, opened 2017-03-21
Add STM32F1 HW support

tekka007:
STM32F103 support, based on #727
- modification eeprom handling
- unique hwID
- rfm69 driver: select spi divider based on cpu freq
- tested rf24 and rfm69 radios


tekka007:

knopserl:
@tekka007 as I still have problems with ESP32 and sleep (it is not implemented and returns) and I have ram and flash size problems with the ATMEGA328p, I'm wondering if the STM32 support works with the proper sleep modes (in contrats to the ESP32) and I can use if for battery powered sensors and my ePaper (which consumes too much memory RAM/Flash of the miniPro)

PR #798, opened 2017-03-24
Clear RX flag in RF24_irqHandler

bathoryr:
Fix for using IRQ pin with NRF24. This should fix issue when gateway or node stops receiving packets after some time.
Tested on Raspberry Pi gateway with --my-rf24-irq-pin option and with repeater node on Nano.

### DIFF
--- a/drivers/RF24/RF24.cpp
+++ b/drivers/RF24/RF24.cpp
@@ -357,8 +357,13 @@ LOCAL void RF24_irqHandler(void)
 
 		// Start checking if RX-FIFO is not empty, as we might end up here from an interrupt
 		// for a message we've already read.
-		while (RF24_isDataAvailable()) {
-			RF24_receiveCallback();		// Must call RF24_readMessage(), which will clear RX_DR IRQ !
+		if (RF24_isDataAvailable()) {
+			do {
+				RF24_receiveCallback();		// Must call RF24_readMessage(), which will clear RX_DR IRQ !
+			} while (RF24_isDataAvailable());
+		} else {
+			// Occasionally interrupt is triggered but no data available - clear RX interrupt only
+			RF24_setStatus(_BV(RF24_RX_DR));
 		}
 		// Restore our interrupt handler.
 #if defined(MY_GATEWAY_SERIAL) && !defined(__linux__)


marceloaqno:
Sorry but I don't see how those changes would fix the problem.
If for some reason the gateway misses an interrupt, RF24_irqHandler() will never be called, thus not having the expected result.

marceloaqno:
I was working on a solution to this problem a while back, but for now I only have an ugly fix for the RPi:
https://github.com/marceloaqno/MySensors/commit/30895a8d94a798ed5a158971a43cad6ba1d8cd31

bathoryr:
In my environment, when interrupt routine is called, then sometimes RF24_isDataAvailable() returns false. So function RF24_receiveCallback() was never called in original code.
My RPi gateway is running more three days with this fix ;) 
Big thanks for your effort @marceloaqno !

marceloaqno:
@bathoryr the problem where RF24_isDataAvailable() returns false, does this also happen with nano or just in RPi?

bathoryr:
I'm running patched version on Nano, I'll do more tests on weekend.

bathoryr:
@marceloaqno  I haven't had time yet to debug it precisely, but yes, Nano behaves the same way - stops receiving after some time. I will post more info.

bathoryr:
@marceloaqno Nano repeater is working with or without this patch, even if RF24_isDataAvailable() returns false. Previous instability was caused by too long distance between nodes. RPi gateway (patched version, irq enabled) is running stable for many weeks.

marceloaqno:
@bathoryr Thank you for all your efforts on this.

Yveaux:
@bathoryr I have a setup with ATMega328+nRF24 running as gateway for months now without any lockup issues. This seems to match your experience withe the nano setup.
The nRF should never trigger an RX interrupt without any data being available. When this does seem to happen I suspect either a hardware error, or an issue with the RPi port.
Your suggested solution contains imho a race condition, when a message comes in just before clearing the RX interrupt flag.

@marceloaqno could it be that the RPi port has an issue with nested interrupts or an issue with incoming interrupts while handling another? Is the interrupt pin configured as active low on the RPi?

marceloaqno:
Still running some tests, but #835 seems to be the root of the locking up problem.

Yveaux:
Very good to hear you seem to have found the root cause of this issue!

marceloaqno:
@bathoryr let us know if you are still experiencing this problem on your system with the latest updates.

bathoryr:
Well done, thank you.

Issue #812, opened 2017-04-01
AVR Sleep times deviate from watchdog timer


Yveaux:
For sleep times <= 125ms the breakdown into individual watchdog timeout intervals is wrong.
This might cause deviation of actual sleep time from requested sleep time.

Issue #811, opened 2017-04-01
AVR might wake from pending external interrupt


Yveaux:
Pending external interrupts are not cleared when attaching an interrupt handler.
This might cause an AVR to immediately wake up again when going to sleep, without an interrupt actually triggering the wake.
Ref: https://forum.arduino.cc/index.php?topic=59217.0

Issue #809, opened 2017-03-30
AVR might sleep with interrupts disabled


Yveaux:
See potential issue here: http://playground.arduino.cc/Learning/ArduinoSleepCode

MySensors interrupt handlers for sleep (wakeUp1() & wakeUp2()) should disable sleep, by adding a call to sleep_disable()

Issue #528, opened 2016-07-29
Problem with sleep/interrupts w/atmega328p


mannkind:
I have a node that sleeps until interrupted on digital pin 2, sends the state of the digital pin 2, then goes back to sleep until interrupted again.  I'm utilizing the MySensors `sleep(INTERRUPT, CHANGE)` function at the end of the `loop()`

Most recently, I was using commit 312a69b and everything was working as expected. Tonight, because I was bored, I started to upgrade to 929b0013 (the latest commit in development as of writing this). Immediately I noticed an issue – the node was no longer sleeping until interrupt and was rapidly sending the state of the pin over and over and over again. 

Checking the commit history, I see there are two commits related to interrupts: 1305aec and 83cbb92. The problem seems to have been introduced in commit 1305aec. I can revert from 312a69b - f0f2542d (the one just before 1305aec) and the issue goes away.

Yveaux:
Main change in this changeset is level changes on the interrupt pin could possible be missed in teh old code. The changes suppress interrupts until going to sleep. If an interrupt happens inside the sleep() function call it should now be detected at all times.
Could it be that in tyour case the interrupt fires immediately when trying to go to sleep (that would previously be ignored)?
Can you provide an exact description of your test setup; e.g. what is used to trigger the interrupt etc.?

mannkind:
The interrupt is physically triggered by a normally open w/magnet reed switch - magnet present and the circuit is open, magnet taken away the circuit is closed. 

The pin is pulled high in `setup()`

In case the reed switch was coincidentally going bad/triggering the interrupt like you mentioned, I removed it from the circuit and used jumper-wire to close/open the circuit.  The problem still persists.

mannkind:
This sketch (based on one of the examples) replicates the problem.
```
// Based on https://raw.githubusercontent.com/mysensors/MySensors/development/examples/BinarySwitchSleepSensor/BinarySwitchSleepSensor.ino
#define MY_RADIO_NRF24
#include <SPI.h>
#include <MySensors.h>

MyMessage msg(0, V_TRIPPED);

void setup() {
  pinMode(2, INPUT_PULLUP);
}

void presentation() {
  present(0, S_DOOR);
}

void loop() {
  uint8_t value = digitalRead(2);
  send(msg.set(value==HIGH));

  sleep(0, CHANGE);
}

Yveaux:
[...]

PR #813, opened 2017-04-01
AVR sleep with external interrupts fixes

Yveaux:
Fixing #812, #811, #809


tekka007:

mannkind:
Huzzah! For what it's worth, this seems to have fixed the issue with sleep & interrupts that I've been having since last summer (#528) 👍

Yveaux:
@mannkind Thanks for the feedback! Very helpful as this issue was very hard to even reproduce and I need user feedback to know if all issues are definately fixed now. 👍

PR #814, opened 2017-04-02
Fix watchdog sleep calculation

Yveaux:
Round requested sleep time up to next multiple of 16ms, to sleep at
least requested time.
Sleep time of 0 will return immediately.
Fixed watchdog sleep time calculation. AVR libc watchdog time
implementation is inconsistent, and requires larger flash/ram usage.


tekka007:

PR #815, opened 2017-04-03
Sleep tweak: clear interrupt flag based on the interrupt used.

mannkind:
This is a small tweak to the last sleep fix. The last fix would clear the proper interrupt flag assuming interrupt1 is INT0 and interrupt2 is INT1. If interrupt1 is INT1 or interrupt2 is INT0, the sleep bug is still triggered.

e.g. `sleep(digitalPinToInterrupt(2), CHANGE, 0);` sleep bug is fixed
 `sleep(digitalPinToInterrupt(3), CHANGE, 0)` still triggers the sleep bug

CC: @Yveaux

### DIFF
--- a/core/MyHwAVR.cpp
+++ b/core/MyHwAVR.cpp
@@ -163,11 +163,11 @@ int8_t hwSleep(uint8_t interrupt1, uint8_t mode1, uint8_t interrupt2, uint8_t mo
 	// to prevent waking immediately again.
 	// Ref: https://forum.arduino.cc/index.php?topic=59217.0
 	if (interrupt1 != INVALID_INTERRUPT_NUM) {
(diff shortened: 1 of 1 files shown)


Yveaux:
Good find!
I'm not sure if this works for all avr devices, but I'll merge it for now. 
Thanks!

mannkind:
Thanks! It is indeed not good enough for all the supported AVR devices, but it's a start. :) That and I only have various atmega328p devices available for testing.

Yveaux:
I changed your fix into a generic avr implementation: #816

mannkind:
I now see that the definition of INTF# seems to be the #, so that makes it easy. I'm glad to see a generic fix. 

Thanks again for your help solving that annoying sleep issue.

PR #816, opened 2017-04-03
Fix clear pending ints for AVR

Yveaux:
Support full AVR range, not only the ones supporting external interrupts 0 & 1.
Only tested on ATMega328 hardware however, as I have no other AVRs available ;-(


Yveaux:

PR #821, opened 2017-04-10
Corrections for stm32f1 serial communication

freynder:
I experienced some issues trying to use Serial1 on a STM32F103C8. This PR should address these:

- use MY_SERIALDEVICE in hwInit
- disable use of MY_SERIALDEVICE.flush() since the current STM32duino implementation performs a reset instead of an actual flush:

### DIFF
--- a/core/MyHwSTM32F1.cpp
+++ b/core/MyHwSTM32F1.cpp
@@ -47,7 +47,7 @@
 bool hwInit(void)
 {
 #if !defined(MY_DISABLED_SERIAL)
-	Serial.begin(MY_BAUD_RATE);
+	MY_SERIALDEVICE.begin(MY_BAUD_RATE);
 #endif
 	if (EEPROM.init() == EEPROM_OK) {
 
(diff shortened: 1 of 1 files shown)


mtiutiu:
Looking into the implementation details about flush method I see this in HardwareSerial.cpp

```

freynder:
> The same is for usart_reset_tx - it just clears the tx buffer. So maybe the reset functions are a little bit misleading named as it doesn't reset the serial port - it just "resets" the buffers(well...flushes them so to speak as it moves the tail pointer to the head).

Hi. I'm not sure what exactly you are saying. Do you mean the implementation is correct? By "flushing" a buffer, I understand that the remaining items in the buffer should be **processed** until empty, not just discarding them. See https://github.com/arduino/Arduino/blob/2bfe164b9a5835e8cb6e194b928538a9093be333/hardware/arduino/avr/cores/arduino/HardwareSerial.cpp#L193 for example.

mtiutiu:
I was only saying that it's misleading as they named it "reset" and you mentioned about reset in this issue. As I'm a more of a hardware guy than a software one when I see "reset" I'm thinking about something else :smile: .There nothing wrong with this PR.

freynder:
I just wanted to make sure I understood correctly. Thanks for commenting on it.

Issue #802, opened 2017-03-28
I2C EEPROM support for OTA using dual Optiboot


kisse66:
Enhancement suggestion:
I need alternate support for OTA using I2C EEPROM instead of the supported SPI Flash. Reason simply being I have boards with those (24C256/512).
I have already implemented experimental changes over development branch through not very well tested yet. Code for MySensors is almost publishable. Optiboot changes need more polishing, but work already.

So question is if there's interest in merging this feature? 

My approach was to make minimal changes to mysensors by implementing I2C EEPROM driver that looks like the SPI flash driver. There's just a few lines plus the new driver, which gets activated by a new compile time def. IF this sounds ok I can prepare it for PR.

tekka007:
@kisse66 Sounds good, please go ahead and submit your PR 👍

Tokay-tech:
Hello,
Are you still working on this enhancement?

kisse66:
Sorry for the delay. Got some personal issues and hobbies progressing slowly. I'll try to do it asap.
Code itself seems to work, but not very well tested so far. Also need to refresh the fork.

kisse66:
Updated the fork and re-tested, but now I get CRC failure after all blocks are received. Need to dig a bit deeper, can also be an issue with my board (cheap EEPROM, unfortunately soldered SOIC).
I'm testing it with arduino pro mini / nrf24. GW running on Raspberry, using MYScontroller 1.0.0.beta. This used to work. Well, tested like 3 or 4 times.

If someone wants to try the fork as-is it's now at https://github.com/kisse66/MySensors.git
Bootloader is not yet there, so one can only see if the EEPROM is handled ok. Need to define MY_OTA_USE_I2C_EEPROM, which disables SPI and uses I2C. HW I2C assumed.

kisse66:
OK, seems to work now. Just a stupid typo.
I modified the code to use extEEPROM instead of duplicating the functionality.

PR #834, opened 2017-05-08
I2C EEPROM support for OTA using dual Optiboot #802

kisse66:
Adds support for I2C EEPROM (>=256kb) OTA. Introduces a wrapper driver (I2CEEPROM) over extEEPROM to emulate SPI  flash, so changes to MySensors is minimal. Tested on Arduino pro mini 8MHz with 24C256/512.


kisse66:
Trying to change history by modifying commit messages (reword) ended up in some conflict. Oldest commit to change is from November and there a lot of changes since due to merging upstream a few times.
IS it necessary to just change commit messages to start with a capitol letter as I understood the gitler check result?

henrikekblad:
It should be possible to amend your previous commits (remove the one where you modify README.md which makes no sense) and change the commit message of  "compiler warning fix" when force push it back to your fork.

kisse66:
yeah, wasn't really thinking PR when forking. I'll try again.

kisse66:
Seems the upstream syncs (merges) I've done along the way complicate it. I can easily edit or remove any commit after the two merges I made, but anything before those (like the readme edit) will break it, i.e. stops to a conflict of something I haven't touched. If I add --preserve-merges I get different set of problems. There was one conflict I had to earlier solve manually (OTA code).

Other issues are fine now, but can't get rid of the old readme mod. So seems I need to google and try some more.

kisse66:
Have tested this a few times and it seems to work. If someone wishes to test there is now forked bootloader also at https://github.com/kisse66/DualOptiboot.git
Targets atmega328_e and _e8 (for 8MHz) build for I2C.

This loader does not fit into 1kB (1330B currently), so fuses must be set for 1024W boot block.
It uses 57600 bps for serial download as default, the same as standard pro mini does.

tekka007:
@kisse66  That's an interesting PR - thanks for submitting 👍 
 Could you possibly change this PR to use the included I2C eeprom library found here: https://github.com/mysensors/MySensors/tree/development/drivers/extEEPROM

kisse66:
a bit confused as it is using extEEPROM. The early one did not, but then I discovered there already was driver and made it use that. Ah ok, the extra EEPROM driver. It is a stub layer or interface modifier (maybe it needs a better class name like IC2Stub?) to make extEEPROM look exactly like the existing Flash driver. This way I only needed to modify one line of mysensors OTA core (creating a different instance) instead of a lot more changes in many places.  If you'd like to remove the extra I2CEeprom class it needs a lot more changes and ifdefs in the OTA code since the driver APIs are different.  I can change of course but not sure how it would look like yet. Did you have something specific in mind ?

kisse66:
'Shim' was the term I was after above, not stub as it's a wrapper. What if I rename it to I2CShim or I2CWrapper?  Then it's a question maybe also where should it be located as it's not a driver. hal/ does not look right to me.  Or would you still prefer other approach?
A bigger and perhaps cleaner approach would be to create a totally new driver interface for storage based OTA. That  would then adapt to the configured actual storage, SPI flash, IC2 EEPROM, ...
That would be so big change, that I'm not sure if I have time to do that and not break anything existing.

fallberg:
Gitler, check this please

fallberg:
A rebase probably resolves the Gitler complaints.

kisse66:
great, somehow I screwed up git update. Old commits already cleaned up re-appeared.  Need to figure out what happened

PR #851, opened 2017-05-20
Minor updates STM32F1 support

tekka007:


hal/architecture/MyHwSTM32F1.cpp
```diff
@@ -120,6 +117,31 @@ int8_t hwSleep(uint8_t interrupt1, uint8_t mode1, uint8_t interrupt2, uint8_t mo
...
+
+void hwRandomNumberInit(void)
+{
+	// use internal temperature sensor as noise source
+	adc_reg_map *regs = ADC1->regs;
+	regs->CR2 |= ADC_CR2_TSVREFE;
+	regs->SMPR1 |= ADC_SMPR1_SMP16;
+
+	uint32_t seed = 0;
+	uint16_t currentValue, newValue = 0;
```
d00616: I think the values needs to be initialized with adc_read(ADC1, 16). If not line 135 is true with every first run.

tekka007: that's ok and will not affect the randomness, since it's adc-derived

PR #866, opened 2017-05-27
Fix ext. EEPROM update()

tekka007:
Fix: EEPROM::update()
- infinite loop when nBytes > 255


tekka007:

PR #873, opened 2017-06-05
Let signing backend validate personalization

### LINKED ISSUE #872: chance to save a big chunk memory with signing
Hi,

as I keep running out of "dynamic memory" in quite a few node sketches, I took a look at whats using the memory.

```
avr-readelf -a /tmp/arduino_build_923073/node-light-tsl2951.ino.elf  | grep 'OBJ.* 3 ' |  sort -nk3
...
   360: 0080038f    33 OBJECT  GLOBAL HIDDEN     3 _msg
   433: 0080036e    33 OBJECT  GLOBAL HIDDEN     3 _msgTmp
   455: 008003b0    33 OBJECT  GLOBAL HIDDEN     3 _msgSign
   341: 00800445    42 OBJECT  GLOBAL HIDDEN     3 _ZL24_signing_verifying_n
   386: 00800416    42 OBJECT  GLOBAL HIDDEN     3 _ZL22_signing_signing_non
   406: 00800285    51 OBJECT  GLOBAL HIDDEN     3 _convBuf
   422: 008002bc   157 OBJECT  GLOBAL HIDDEN     3 Serial
   536: 008005b8   197 OBJECT  GLOBAL HIDDEN     3 _soft_sha256
   557: 0080046f   197 OBJECT  GLOBAL HIDDEN     3 _signing_sha256
```

_signing_sha256 is the Sha256Class used by the software signing backed. Ok. That's the price for not having hardware.

But if I look at _soft_sha256 in MySigning.cpp, I see it just being used for the EEPROM checksum validation... and the result is then ignored!!! by signerInit(). Wasting (more than) 197 bytes for a checksum that's ignored sounds questionable.

Actually, the checksum just seems to be verified within the SecurityPersonalizer sketch (... that's used to set the EEPROM values).

Removing _soft_sha256 (and related code) frees 229 Bytes and the node keeps working... it's just the Personalizer that needs it.

Though, just removing it sounds bad... and replacing it with some other Checksum will make the current EEPROM checksum invalid.

Are you interested in a patch/pull request to remove it? Or do you prefer a patch to defined the Sha256Class instances locally in signerValidatePersonalization() and pass the reference to signerSha256Init/-Update/-Final() (likely introduce memory issues in other situations)?

This is with the following defines:
```
#define MY_RADIO_NRF24

// signing (would use +600 bytes RAM with atsha)
#define MY_SIGNING_REQUEST_SIGNATURES
#define MY_SIGNING_GW_REQUEST_SIGNATURES_FROM_ALL
// ... but I use sw backend (+ another 100 bytes RAM)
#define MY_SIGNING_SOFT
```

fallberg:
For soft signing, let backend handle checksum verification of
personalization data. This ensures that the backends sha256 library is
used and avoids an unnecessary instanitation of the public library.
For atsha204 signing, no checksum verification is made, just chip
locking flags.

Fixes #872

### DIFF
--- a/core/MySigning.cpp
+++ b/core/MySigning.cpp
@@ -90,7 +90,6 @@ extern bool signerAtsha204SignMsg(MyMessage &msg);
 #define signerBackendVerifyMsg  signerAtsha204VerifyMsg
 #define signerBackendSignMsg    signerAtsha204SignMsg
 #endif
-static bool signerValidatePersonalization(void);
 static bool skipSign(MyMessage &msg);
 #else // not MY_SIGNING_FEATURE
 #define signerBackendCheckTimer() true
@@ -103,8 +102,6 @@ static bool signerInternalProcessNonceResponse(MyMessage &msg);
 void signerInit(void)
 {
 #if defined(MY_SIGNING_FEATURE)
-	(void)signerValidatePersonalization();
-
 	// Read out the signing requirements from EEPROM
 	hwReadConfigBlock((void*)_doSign, (void*)EEPROM_SIGNING_REQUIREMENT_TABLE_ADDRESS,
 	                  sizeof(_doSign));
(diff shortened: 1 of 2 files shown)


fallberg:

PR #874, opened 2017-06-06
memory optimizations in signing

rclasen:
Hi,

due to memory constraints in some of my node sketches I looked for what's actually using a lot of memory and found some of the bits in the signing code to use some quite "big" chunks. Well... "big" in relation to the 2k total an Arduino has.

While my current need is addressed by #872 and #873, Patrick suggested that upcoming changes might reintroduce the extra memory consumption.

So, I looked for further chances to reduce the memory usage with the public signing API being used.

The individual Ideas each have their own commit... and most notably the last commit likely isn't acceptable.

This works for me (tm)... but I might be missing bits...

Please comment if there is anything I shall change/cleanup/... or if none of the ideas are acceptable ;)

Rainer

### DIFF
--- a/core/MySigning.cpp
+++ b/core/MySigning.cpp
@@ -284,12 +284,10 @@ bool signerVerifyMsg(MyMessage &msg)
 	return verificationResult;
 }
 
-static uint8_t sha256_hash[32];
 Sha256Class _soft_sha256;
 
 void signerSha256Init(void)
 {
-	memset(sha256_hash, 0, 32);
 	_soft_sha256.init();
 }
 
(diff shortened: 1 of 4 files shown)


fallberg:
Will this change still permit the use of the public hash API to be used independently of the one in the software backend? They have to be separated. That is, ongoing hash operations in the backend must not interfere with ongoing operations through the public API.
I suppose it would work as they are two separate instances, but please confirm this.

rclasen:
Hi,

yeah, thats what I was expecting... so... removed this bit. The rest should be fine and completely non-intrusive.

Rainer

fallberg:
I agree. Nice job and thanks for the fixes.

PR #875, opened 2017-06-06
Add progmem support for String messages

rclasen:
added variant of MyMessage::set and sendSketchInfo that accept a string that's stored in PROGMEM.

This avoids the need to (temporarily) store strings in memory to get them into a message and thereby saves memory.

I'm also using this to send literal error messages.

I haven't found any real cases in the MySensors tree that are using the __FlashStringHelper. If a variant similar to other progmem related functions using set_P(...) and sendSketchinfo_P(...) is preferred, please let me know.

### DIFF
--- a/core/MyMessage.cpp
+++ b/core/MyMessage.cpp
@@ -249,6 +249,23 @@ MyMessage& MyMessage::set(const char* value)
 	return *this;
 }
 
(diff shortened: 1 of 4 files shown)


tekka007:

PR #877, opened 2017-06-08
Keep Serial from being linked when disabled

rclasen:
Even with MY_DISABLED_SERIAL defined, the linker is still adding the Serial
object to the binary.

That's because a) the MySensors *.cpp files are actually included into
the sketch and compiled as part of it (and not compiled separately and
then linked) and b) due the the default(?) behavior of the linker
(https://github.com/arduino/Arduino/issues/4579).

### DIFF
--- a/hal/architecture/MyHwAVR.cpp
+++ b/hal/architecture/MyHwAVR.cpp
@@ -303,6 +303,7 @@ uint16_t hwFreeMem()
 
 void hwDebugPrint(const char *fmt, ... )
 {
+#ifndef MY_DISABLED_SERIAL
 	char fmtBuffer[MY_SERIAL_OUTPUT_SIZE];
 #ifdef MY_GATEWAY_FEATURE
 	// prepend debug message to be handled correctly by controller (C_INTERNAL, I_LOG_MESSAGE)
@@ -327,4 +328,5 @@ void hwDebugPrint(const char *fmt, ... )
 	va_end (args);
 	MY_SERIALDEVICE.print(fmtBuffer);
 	MY_SERIALDEVICE.flush();
+#endif
 }
--- a/hal/architecture/MyHwESP8266.cpp
+++ b/hal/architecture/MyHwESP8266.cpp
@@ -149,6 +149,7 @@ uint16_t hwFreeMem()
 
 void hwDebugPrint(const char *fmt, ... )
 {
+#ifndef MY_DISABLED_SERIAL
 	char fmtBuffer[MY_SERIAL_OUTPUT_SIZE];
 #ifdef MY_GATEWAY_FEATURE
 	// prepend debug message to be handled correctly by controller (C_INTERNAL, I_LOG_MESSAGE)
@@ -173,4 +174,5 @@ void hwDebugPrint(const char *fmt, ... )
 	va_end (args);
 	MY_SERIALDEVICE.print(fmtBuffer);
 	MY_SERIALDEVICE.flush();
+#endif
 }
--- a/hal/architecture/MyHwNRF5.cpp
+++ b/hal/architecture/MyHwNRF5.cpp
@@ -472,6 +472,7 @@ uint16_t hwFreeMem()
 
 void hwDebugPrint(const char *fmt, ...)
 {
+#ifndef MY_DISABLED_SERIAL
 	char fmtBuffer[MY_SERIAL_OUTPUT_SIZE];
 #ifdef MY_GATEWAY_FEATURE
 	// prepend debug message to be handled correctly by controller (C_INTERNAL, I_LOG_MESSAGE)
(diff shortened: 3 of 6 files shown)


fallberg:
What about other architectures?

rclasen:
Was afraid to touch the other architectures as I have no chance to test them.
Though, will add a commit and look at jenkins results.

rclasen:
As linux doesn't use a serial and creative defines may result in MY_DISABLED_SERIAL and DEBUG_OUTPUT_ENABLED (used for the prototype in MyHw.h) being defined, I changed the implementation to keep hwDebugPrint() - just without Serial code if MY_DISABLED_SERIAL is set.

rclasen:
Hmm, now that I think of it... would it be better to suppress the whole function (and not just the body) if DEBUG_OUTPUT_ENABLED isn't set? This would match the handling of the prototype in MyHw.h - while MY_DISABLED_SERIAL might not be sufficient if user is messing up his defines (eg. manually setting any MY_DEBUG_VERBOSE_*).

tekka007:
@rclasen Just ran a few tests - see below:

This is my setup:
- Arduino IDE 1.8.3
- AVR board defs 1.6.19
- [RelayActuator](https://github.com/mysensors/MySensors/blob/development/examples/RelayActuator/RelayActuator.ino) sketch (from the MySensors examples)
- Arduino UNO

Current development branch ([59337b1](https://github.com/mysensors/MySensors/commit/59337b19fe2280e3d375a5e70156c75379d18196)):
using ```#define MY_DEBUG```:
FLASH 14748 bytes
RAM 491 bytes

using ```#define MY_DISABLED_SERIAL```:
FLASH 8064 bytes
RAM 422 bytes

Using this PR ([113af1f ](https://github.com/mysensors/MySensors/pull/877/commits/113af1f77d41347b855f3c5affa6ec1e7ba955bc))

using ```#define MY_DEBUG``` :
FLASH 14748 bytes
RAM 491 bytes

using ```#define MY_DISABLED_SERIAL```:
FLASH 8064 bytes
RAM 422 bytes

So, same results with dev branch and this PR - am I missing something?
What are your findings?

rclasen:
the RelayActor is using Serial by itself... which causes the linker to pull in the Serial object.
Please remove the Serial.print(ln) lines in receive() to compare the memory Usage.

For more details (=what is taking the memory), I tend to look at:

`avr-readelf -a /tmp/arduino_build_*/*.elf  | grep 'OBJ.* 3 ' | sort -nk3`

So, when looking at the RelayActor sketch with it's own Serial it looks like this:

Without MY_DISABLED_SERIAL I'm getting:

```
Global variables use 420 bytes (20%) of dynamic memory, leaving 1628 bytes for local variables. Maximum is 2048 bytes.

klick:MySensors$ avr-readelf -a /tmp/arduino_build_8*/*.elf  | grep 'OBJ.* 3 ' | sort -nk3 | tail -n 3
   120: 0080027a    33 OBJECT  GLOBAL HIDDEN     3 _msg
   153: 00800259    33 OBJECT  GLOBAL HIDDEN     3 _msgTmp
   146: 008001bc   157 OBJECT  GLOBAL HIDDEN     3 Serial
```

with MY_DISABLED_SERIAL and my PR I get:
```
Global variables use 201 bytes (9%) of dynamic memory, leaving 1847 bytes for local variables. Maximum is 2048 bytes.

klick:MySensors$ avr-readelf -a /tmp/arduino_build_8*/*.elf  | grep 'OBJ.* 3 ' | sort -nk3 | tail -n 3
   126: 00800165    15 OBJECT  GLOBAL HIDDEN     3 _ZL12_transportSM.lto_pri
   109: 0080017a    33 OBJECT  GLOBAL HIDDEN     3 _msg
   143: 0080019b    33 OBJECT  GLOBAL HIDDEN     3 _msgTmp


```

tekka007:
@rclasen Of course! Totally overlooked that - well, that's a huge difference 👍

tekka007:

rclasen:
hmm, actually... my PR sucks... the Serial object shouldn't be linked in when MY_DEBUG (and related) aren't set, aswell.

Having to define MY_DISABLED_SERIAL **in addition** to actually get rid of it is suboptimal.

rclasen:
ok, tried to use DEBUG_OUTPUT_ENABLED instead of MY_DISABLED_SERIAL ... but then I still need the later to get rid of the references in hwInit and hwPowerDown (... on AVR). So... the PR as is doesn't seem that bad ;)

tekka007:
It's up to the user to initialise and flush serial output if MY_DEBUG (and related) is disabled...hence DEBUG_OUTPUT_ENABLED  can be used im hwInit() and hwPowerDown()...

rclasen:
Yeah, once I looked at the preprocessor output with MY_DEBUG and MY_DISABLED_SERIAL both undefined, this became obvious ;)
(was to lazy to do this before).

fallberg:
Gitler, check this please

fallberg:
A rebase probably resolves the Gitler complaints.

PR #888, opened 2017-07-05
Use bool instead of boolean in drivers

mfalkvidd:
In the Arduino IDE, boolean is an alias for bool.
The Arduino IDE might deprecate boolean in the future. See
https://github.com/arduino/Arduino/issues/4673 for details.

Command:
find drivers -type f | xargs sed -i 's/boolean/bool/g'


fallberg:

PR #890, opened 2017-07-06
Fixed indentations using astyle

fallberg:


marceloaqno:
By default the Arduino IDE uses space for indentation. Could an exception be add to the *.ino files?

fallberg:
@henrikekblad you took the decision for tab spacing, what do you say about this? I am all for using spaces all over but I am not that fond of having a mixed setup.

fallberg:
Ideally, the arduino IDE should adhere to the .editorconfig

PR #899, opened 2017-07-26
Added posibility to use non standard SPI Flash

bilbolodz:
Added possibility to tune flash commands and parameters
in case off use non standard SPI flash for OTA

### DIFF
--- a/drivers/SPIFlash/SPIFlash.h
+++ b/drivers/SPIFlash/SPIFlash.h
@@ -75,29 +75,74 @@
 
 #include <SPI.h>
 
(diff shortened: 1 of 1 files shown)


tbowmo:
If using a non standard SPI flash, would that imply changing the bootloader as well?

bilbolodz:
I'm aware of these. It has been already done. But without these patch it's very easy "brick the device" (turn off possibility of OTA).

bilbolodz:
And these patch doesn't "break or force" anything. It's only possibility to use without need of "manual patching" local sources.

PR #912, opened 2017-08-20
Use MY_SERIALDEVICE instead of Serial in examples

mfalkvidd:
MY_SERIALDEVICE is set (by the hardware abstraction layer)
to the appropriate device depending on which board is used.

Most boards use Serial, but SAMD doesn't, so to make the
examples work, MY_SERIALDEVICE should be used.

Command:
find examples -type f | xargs sed -i 's/Serial\.print/MY_SERIALDEVICE\.print/g'


mfalkvidd:
Using Serial is the arduino way, so we should continue using Serial. SAMD already has a macro that makes Serial available.
Closing.

PR #920, opened 2017-09-03
Add numerical library versioning

tekka007:
- alphanumerical (SemVer) and numerical library version representation

### DIFF
--- a/core/Version.h
+++ b/core/Version.h
@@ -16,14 +16,50 @@
  * modify it under the terms of the GNU General Public License
  * version 2 as published by the Free Software Foundation.
  */
-/***
- *  This file defines the Sensor library version number
- *  Normally, contributors should not modify this directly
- *  as it is managed by the MySensors Bot.
- */
+
+/**
+* @file Version.h
+*
+* @defgroup Versiongrp Version
+* @ingroup internals
+* @{
+*
+* This file defines the MySensors library version number
+* Please adjust for new releases.
+
+* These helper macros generate a numerical and alphanumerical (see http://www.semver.org) representation of the library version number, i.e
+*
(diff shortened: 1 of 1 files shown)


tekka007:

core/Version.h
```diff
@@ -16,14 +16,50 @@
...
+*
+* This file defines the MySensors library version number
+* Please adjust for new releases.
+
+* These helper macros generate a numerical and alphanumerical (see http://www.semver.org) representation of the library version number, i.e
+*
+* | SemVer      | Numerical
+* |-------------|-------------
+* | 2.1.0       | 0x020100FF
+* | 2.1.1-beta  | 0x02010101
```
fallberg: is alfa 0x02010100?

fallberg:
One could consider to automatically bump patch version on every submitted PR. That obviously leads to a limited lifetime between releases with 255 patches with this scheme, but maybe that isn't a bad idea?

core/Version.h
tekka007: Yes, it should be an incremental counter for pre-releases, i.e. first pre-release gets 0x00 (I have to fix this): 2.2.0-beta => 0x02020200 (since it is the first pre-release)
Generally, I like the idea of automatically bumping the patch version for every PR - this would be a github hook?

fallberg: Not sure if github supports that. Will have to check up on that. Perhaps jenkins could do it on command like triggering a "submit pull request" order from a github comment by an authorized user. Jenkins could then either amend in a version bump with the merge or create a dedicated commit for bumping version.

fallberg:
Looks good to me. We can consider the automatic bumping later on.

PR #921, opened 2017-09-03
Update hwUniqueID() function

tekka007:
- add HW specific params for AVR (still, no unique ID)
- update padding bytes for ESP8266, Teensy


tekka007:

PR #923, opened 2017-09-05
Fix sanity check feature

tekka007:
Refactoring MyConfig.h resulted in continuously disabled sanity check


tekka007:

Issue #930, opened 2017-09-18
Splash Screen can break backward compatibility


LemarinelNet:
The splash-screen (included since #889) which is set ENABLED by default breaks the compatibility on my openhab binding:

It seems that the bindings always split the incoming frame with semi-column, but the splash lines doesn't contains the required semi-columns, and then the binding code (which is not robust enough) fails in a java unhandled error.

Would'nt it be more secure to output the splash-screen as a common frame (i.e. a comment or debug frame) ? 
Or at least, a more backward-compatible would be to set splash_screen DISABLED by default ?

henrikekblad:
@tekka007, maybe the splash should be printed like normal debug messages on gateway-nodes so it doesn't interfere with the normal command parsing?

LemarinelNet:
PS : @tekka007,Sorry for this off-topic, I've dropped you a message in mysensors forum about OTA ;) Have you seen it ?

tekka007:
@hek This will most likely result in a distorted splash screen (due to the timer field).
I see two options: 
1. disable it if plugin has issues:
`#define MY_SPLASH_SCREEN_DISABLED
`
2. disable it on all GW nodes by default

LemarinelNet:
I vote for a disabled by default behavior.
Having this enabled by default might get many users in trouble if they doesn't understand what's happening ;)

henrikekblad:
@tekka, option 2 is probably best for backward compatibility.

tekka007:
Fixed in #931

PR #931, opened 2017-09-18
Disable splash screen on GW

tekka007:
see #930

### DIFF
--- a/core/MySensorsCore.cpp
+++ b/core/MySensorsCore.cpp
@@ -98,7 +98,7 @@ void _begin(void)
 
 	const bool hwInitResult = hwInit();
 
-#if !defined(MY_SPLASH_SCREEN_DISABLED)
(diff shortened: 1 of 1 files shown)


core/MySensorsCore.cpp
```diff
@@ -98,7 +98,7 @@ void _begin(void)
 
 	const bool hwInitResult = hwInit();
 
-#if !defined(MY_SPLASH_SCREEN_DISABLED)
+#if !defined(MY_SPLASH_SCREEN_DISABLED) & !defined(MY_GATEWAY_FEATURE)
```
henrikekblad: & => &&

tekka007: I need to clean my glasses :) Thanks

henrikekblad: 🤓

PR #939, opened 2017-10-01
Fix uninitialized state handling

d00616:


d00616:

PR #943, opened 2017-10-01
NRF5 fix 8 bit inttype handling

d00616:
Without this commit, formatted outputs of 8 bit values dosn't work with arduino-nrf5 -> https://github.com/sandeepmistry/arduino-nRF5/issues/197


d00616:

PR #973, opened 2017-11-05
Update RelayActuator example

mfalkvidd:
Based on feedback from skywatch,
https://forum.mysensors.org/post/78896

We'll need to update the wiring instructions on
https://www.mysensors.org/build/relay as well.


mfalkvidd:
https://www.mysensors.org/build/relay has been updated with the new wiring instructions

PR #982, opened 2017-11-09
Fix RFM95 reset

tekka007:
RFM95 reset: pull reset pin low for 100uS, release, wait 5ms
see [datasheet](http://www.hoperf.com/upload/rf/RFM95_96_97_98W.pdf), chapter 7.2.2


mfalkvidd:
I've verified that the datasheet says to go low, and checked that the commit history does not indicate a reason for doing otherwise.
I also checked that this is the only place where the RFM95 radio is reset, so there shouldn't be a need to change this somewhere else.
I have verified that the RFM69 specifies the reset signal to be high, so there is no need to change the RFM69 driver.
The difference between RFM69 and RFM95 was probably what caused this confusion initially.

PR #1001, opened 2017-11-26
Add MY_DEBUGDEVICE to redirect debug output

nelsonov:
MY_DEBUGDEVICE, if defined, will replace MY_SERIALDEVICE
for the purpose of printing debug messages. This only
applies to debugging.

As discussed in the forum topic: MY_SERIALDEVICE overrides MY_DEBUG_HWSERIAL

### DIFF
--- a/hal/architecture/AVR/MyHwAVR.cpp
+++ b/hal/architecture/AVR/MyHwAVR.cpp
@@ -313,17 +313,20 @@ uint16_t hwFreeMem()
 
 void hwDebugPrint(const char *fmt, ... )
 {
+#ifndef MY_DEBUGDEVICE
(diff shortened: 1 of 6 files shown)


mfalkvidd:
Great work!
Some things I'd like to see to make this more accessible to users (I can help with this if you want)

- [ ] add to keywords.txt to get syntax highlightning in the Arduino IDE
- [ ] add to MyConfig.h with doxygen comments so this feature gets added to the [documentation](https://www.mysensors.org/apidocs-beta/group__SerialDebugGrpPub.html#ga7e7232f278193dd8684f973ee70ef3f5). Use the [existing debug documentation](https://github.com/mysensors/MySensors/blob/aabddc841cd526ffd716f4e467b4164b2e267a95/MyConfig.h#L32) as template.

PR #1002, opened 2017-11-26
Harmonize debug outputs

tekka007:


examples/GatewayESP8266OTA/GatewayESP8266OTA.ino
```diff
@@ -133,32 +131,33 @@
...
 
+#include <ArduinoOTA.h>
 #include <MySensors.h>
 
 void setup()
 {
 	// Setup locally attached sensors
 	ArduinoOTA.onStart([]() {
-		debug("ArduinoOTA start\n");
+		Serial.println("ArduinoOTA start");
```
mfalkvidd: How come Serial.println is used here instead of GATEWAY_DEBUG ?

tekka007: GATEWAY_DEBUG should not be used in examples

PR #1009, opened 2017-12-03
Extended gateway

pgollor:
I am use the `mysgw` on my Raspberry Pi with MQTT. My smart home system use MySensors throw the MQTT gateway. So in this case it was not possible to use the `requestTime` function, because no controller is connected directly.
So i added a second linux example called `mysgw-ext` (MySensors gateway extended) which answer the time requests. I hope you will find it usable?
It was tested with this test sketch on an Arduino Nano as node: https://gitbucket.pgollor.de/Arduino/timeRequest


pgollor:
It is so time-consuming to change all commits to pass your rules and read the license on clahub for this changes.

Issue #954, opened 2017-10-26
RS485 Serial Gateway - no reception after receive collision 


seeers:
Hi, 

i use a Arduino 2560 with the  RS485 serialgateway configuration to receive humidity an temperature values from 12 nodes connected to HWSerial1 by a Max485. Each node sends its measured values every 90 seconds. Sometimes there are collisions, then nothing is received except the local sensors at the gateway. 

I have recorded the last "receive freezes" with openhab, here are some examples:

2017-10-25 23:06:42.661 Message from gateway received: 23;3;1;0;1;58.20
2017-10-25 23:06:50.374 Message from gateway received: 0;0;1;0;47;89
2017-10-25 23:06:54.160 Message from gateway received: 35;1;1;0;0;20.69
2017-10-25 23:06:54.210 Message from gateway received: 35;2;1;0;0;20.90
2017-10-25 23:06:54.236 Message from gateway received: 35;3;1;0;1;52.40
2017-10-25 23:07:**09**.461 Message from gateway received: **14**;1;1;0;0;19.69
2017-10-25 23:07:**09**.512 Message from gateway received: **14**;2;1;0;0;20.00
2017-10-25 23:07:**09**.560 Message from gateway received: **14**;3;1;0;1;68.30
2017-10-25 23:07:**09**.587 Message from gateway received: **33**;2;1;0;0;23.80
2017-10-25 23:07:**09**.612 Message from gateway received: **33**;3;1;0;1;57.20
2017-10-25 23:07:18.985 Message from gateway received: 31;1;1;0;0;22.69
.....
2017-10-25 23:08:12.529 Message from gateway received: 23;1;1;0;0;21.81
2017-10-25 23:08:12.559 Message from gateway received: 23;2;1;0;0;21.90
2017-10-25 23:08:12.596 Message from gateway received: 23;3;1;0;1;58.10
2017-10-25 23:08:24.116 Message from gateway received: 35;1;1;0;0;20.69
2017-10-25 23:08:24.177 Message from gateway received: 35;2;1;0;0;20.90
2017-10-25 23:08:24.228 Message from gateway received: 35;3;1;0;1;52.40
2017-10-25 23:08:39.461 Message from gateway received: **14**;1;1;0;0;19.69
2017-10-25 23:08:50.372 Message from gateway received: 0;0;1;0;47;91
2017-10-25 23:09:13.784 Message from gateway received: 0;0;1;0;47;88 
[...]

PR #1013, opened 2017-12-17
Add option to send multiple header ini (#954?)

rejoe2:
Offering the option to set multiple header init. 
forum#7454, post#6
Perhaps changing default also to three could be a good idea...

May also help wrt to this PR: 
https://github.com/mysensors/MySensors/issues/954


mfalkvidd:
@rejoe2 would you mind doing the following, to make this feature easier for users to find and use?
- [ ] add to keywords.txt to get syntax highlightning in the Arduino IDE
- [ ] add to MyConfig.h with doxygen comments so this feature gets added to the [documentation](https://www.mysensors.org/apidocs-beta/group__RS485SettingGrpPub.html#gaa7614b7975eccb4f40036dfd007ad885). Use the [existing rs485 settings](https://github.com/mysensors/MySensors/blob/aabddc841cd526ffd716f4e467b4164b2e267a95/MyConfig.h#L203) as template.

The easiest way would probably be to move the _if !defined_ chunk to MyConfig.h and add doxygen comments.

fallberg:
@rejoe2 as long as you keep adding commits and don't amend your failing commits, gitler will never approve this pr. Instructions for amending is mailed to you (or you can click the details link on the gitler toll gate status here)

rejoe2:
Thanks for the feedback. As I'm just a hobbyist these methods and tools are new to me, I didn't realise my commits made online in the repo had directly gone to the main repo. I'll try to read and understand the documents and then try again.

fallberg:
They go into your development branch on your fork of the main repo. And since you have filed a pull request from your development branch to the development branch of  the main repo, they show up here. Don't worry, the worst thing that can happen is that you break your forked repo, and if this is the only change, you can just remove the fork and create another one and apply the same change again.

keywords.txt
```diff
@@ -177,6 +177,7 @@ MY_RS485_BAUD_RATE	LITERAL1
 MY_RS485_MAX_MESSAGE_LENGTH	LITERAL1
 MY_RS485_HWSERIAL	LITERAL1
 MY_RS485_DE_PIN	LITERAL1
+MY_RS485_SOH_COUNT LITERAL1
```
mfalkvidd: The Arduino IDE is really stupid. It requires a **tab** character between "MY_RS485_SOH_COUNT" and "LITERAL1". At the moment you have a space character. Could you fix this?

mfalkvidd:
@rejoe2 about the Gitler error: Gitler doesn't like human-readable URLs (at least not if they are long). You need to obfuscate the URL by using a URL shortening service like bit.ly or goo.gl. See https://github.com/mysensors/MySensors/commit/a23ab14494d28f90b019ea3c651dd61d50382e63 for how I did it when I stumbled into this problem.

rejoe2:
Thanks for your hints. 
Now, I committed a second, cleaned-up version (#1014) as there seems not to exist the possibility to change remarks.

Thanks a lot also for supporting this cool project!

PR #1014, opened 2017-12-17
Add MY_RS485_SOH_COUNT option

rejoe2:
(2. trial to merge this)
Slight modification of suggestion here just offering the option to be set in the sketch: 

from forum topic 7454 Post 6
Perhaps changing default also to three could be a good idea...

May also help wrt to this PR: #954


fallberg:
Nice!

PR #1016, opened 2017-12-19
RFM69 not going to sleep

gmarti:
Had a problem sleeping rfm69.

Seems the code is not waiting for the mode to change before leaving setRadioMode method.

Please, first PR here, not sure i have done it properly.


gmarti:
My bad, **this is not the bug**

I think it may be the RFM69_OPMODE_SEQUENCER_OFF that make my radio not going to sleep.

Will check that tomorrow

PR #1027, opened 2018-01-09
SignalReport disabled by default

tekka007:
This saves 1kB of flash - enable only if needed (`#define MY_SIGNAL_REPORT_ENABLED`)

### DIFF
--- a/MyConfig.h
+++ b/MyConfig.h
@@ -1102,20 +1102,12 @@
 #endif
 
 /**
- * @def MY_DISABLE_SIGNAL_REPORT
- * @ingroup memorysavings
- * @brief If defined, signal report functionality will be unavailable.
- * @see MY_SIGNAL_REPORT_ENABLED
- */
-/**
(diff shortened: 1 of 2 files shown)


MyConfig.h
```diff
@@ -1102,20 +1102,12 @@
...
- * @brief Enables signal report functionality.
- * @note Enabled by default. This feature adds ~1kB code to the sketch.
- * @see MY_DISABLE_SIGNAL_REPORT
- */
-#ifndef MY_DISABLE_SIGNAL_REPORT
-#define MY_SIGNAL_REPORT_ENABLED
-#endif
+* @def MY_SIGNAL_REPORT_ENABLED
+* @brief Enables signal report functionality.
+* @note Enabled by default. This feature adds ~1kB code to the sketch.
```
mfalkvidd: Is signal report enabled by default or not? Commit message says not enabled, doxygen comment says enabled.

tekka007: good catch - will update the PR. Thanks

PR #1039, opened 2018-01-25
Script release preparations

mfalkvidd:
This automates steps 1-3 of
https://github.com/mysensors/MySensors/wiki/Library-releases

### DIFF
--- a/.mystools/prepare-release.sh
+++ b/.mystools/prepare-release.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+if [ ! -d mysensors-release ]; then
+	git clone -o upstream git@github.com:mysensors/MySensors.git mysensors-release
+fi
+cd mysensors-release || exit
+git fetch upstream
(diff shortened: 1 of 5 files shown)


.mystools/prepare-release.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+if [ ! -d mysensors-release ]; then
+	git clone -o upstream git@github.com:mysensors/MySensors.git mysensors-releas
```
fallberg: typo: mysensors-releas

mfalkvidd: Good catch. I have pushed an update.

fallberg:
With the development branch now properly in sync with master, we hopefully won't have to do this very complicated merging anymore. It should just be a regular merge that generates no conflicts from now on.

mfalkvidd:
Ok. Lets close this and update the wiki instead. Not complicated is good.

Issue #1022, opened 2017-12-29
Option to disable `isatty` check for Linux gateway


ConnorRigby:
I'm developing a controller in Erlang/Elixir, And i would like to use the Raspberry Pi SPI bus directly, rather than using an external Arduino gateway. I've hit a small problem where spawning programs from the Erlang Beam VM fails the `isatty` check, so i don't get logs from `mysgw`.  

https://github.com/mysensors/MySensors/blob/development/hal/architecture/Linux/MyMainLinux.cpp#L415

mfalkvidd:
@ConnorRigby what configure command did you use when configuring mysgw?

Does --my-serial-is-pty help?

mfalkvidd:
I used https://raw.githubusercontent.com/skuhl/sys-prog-examples/master/simple-examples/isatty.c to test the following:
```
pi@raspberrypi:~ $ ./a.out
stdin is a terminal.
stdout is a terminal.
stderr is a terminal.
pi@raspberrypi:~ $ touch /tmp/file
pi@raspberrypi:~ $ ./a.out < /tmp/file
stdin is not a terminal.
stdout is a terminal.
stderr is a terminal.
root@raspberrypi:/home/pi# ./a.out < /dev/ttyUSB42
stdin is a terminal.
stdout is a terminal.
stderr is a terminal.
```

mysgw configuration:
`./configure --my-transport=none --my-gateway=serial --my-serial-is-pty --my-serial-pty=/dev/ttyUSB42`

So it seems like --my-serial-is-pty does the trick.

ConnorRigby:
Will try that out right now!

ConnorRigby:
Unfortunately that still did not work. I'm using the Network Gateway by the way. I thought `--my-serial-is-pty` was only for a `serial` gateway
```bash
./configure \
--c_compiler=/home/connor/.nerves/artifacts/nerves_toolchain_armv6_rpi_linux_gnueabi-0.12.1.linux-x86_64/bin/armv6-rpi-linux-gnueabi-gcc\
--cxx_compiler=/home/connor/.nerves/artifacts/nerves_toolchain_armv6_rpi_linux_gnueabi-0.12.1.linux-x86_64/bin/armv6-rpi-linux-gnueabi-g++ \
--my-transport=nrf24 \
--my-config-file=/tmp/mysensors.dat \
--my-debug=enable\ 
--bindir=/home/connor/oss/home_controller/deleteme/priv/my_sensors \
[...]

PR #1061, opened 2018-02-11
Linux: Use config file for gateway settings

marceloaqno:
- The following settings can be use on the config file:
  - verbose=[debug,info,notice,warn,err] - Logging verbosity. Default=debug
  - log_file=[0|1] - Enable logging to a file. Default=0
  - log_filepath=[/tmp/mysgw.log] - Log file path.
  - log_pipe=[0|1] - Enable logging to a named pipe(aka fifo). Default=0
    Use this option to view your gateway's log messages from the
    log_pipe_file (defined below).
    To do so, run the following command on another terminal:
    - $ cat "log_pipe_file"
  - log_pipe_file=[/tmp/mysgw.pipe]
  - syslog=[0|1] - Enable logging to syslog. Default=0
  - eeprom_file=[/etc/mysensors.eeprom]
  - eeprom_size=[1024]
- Change some mysgw parameters:
  - Added:
    - -q, --quiet:  for quiet mode, disable log messages written to the
    terminal.
  - Removed:
    - -d, --debug: removed, log messages are now enabled by default.
  - Replaced:
    - -b, --background: replaced by --daemon
- isatty() is no longer used, log messages by default are printed to
  stderr unless the gateway is started with --quiet (#1022)
- MY_LINUX_CONFIG_FILE: no longer holds the path to the eeprom file,
  but to the configuration file

### DIFF
--- a/MyConfig.h
+++ b/MyConfig.h
@@ -1852,7 +1852,7 @@
  * @note For now the configuration file is only used to store the emulated eeprom state.
  */
 #ifndef MY_LINUX_CONFIG_FILE
-#define MY_LINUX_CONFIG_FILE "/etc/mysensors.dat"
+#define MY_LINUX_CONFIG_FILE "/etc/mysensors.conf"
 #endif
 /** @}*/ // End of LinuxSettingGrpPub group
 /** @}*/ // End of PlatformSettingGrpPub group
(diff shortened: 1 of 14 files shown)


mfalkvidd:
We'll need to document the possible config file settings and their default values, preferably in Doxygen. Perhaps an activity for the Hackathon?

mfalkvidd:
We'll need to update the build page with these changes when doing the release.

Issue #993, opened 2017-11-15
GatewayESP8266 inclusion button misleading


mfalkvidd:
The code for the ESP8266 WiFi Gateway (https://github.com/mysensors/MySensors/blob/master/examples/GatewayESP8266/GatewayESP8266.ino) is not correct when it comes to the inclusion button. Either the comments on which GPIO the button should be connected to or the PIN assignment should be changed.

Problem found by MySensors user Strixx. See https://forum.mysensors.org/post/79621 for more details.

mfalkvidd:
I tried to verify this today, but was not able to.
I uncommented these from the GatewayESP8266 example:
```
#define MY_INCLUSION_MODE_FEATURE
// Enable Inclusion mode button on gateway
#define MY_INCLUSION_BUTTON_FEATURE
// Set inclusion mode duration (in seconds)
#define MY_INCLUSION_MODE_DURATION 60
// Digital pin used for inclusion mode button
#define MY_INCLUSION_MODE_BUTTON_PIN D1
```
I also commented out the radio choice, so I didn't have to connect a radio,.
Then I connected a dupont cable to D1 on my NodeMCU 1.0 and uploaded the sketch. When briefly (~0.5 seconds) connecting the dupong cable to a GND pin, nothing happens on the serial debug output, and nothing happens on the gateway tcp connection.
Serial debug:
```
5497 MCO:BGN:INIT GW,CP=R-NGE---,VER=2.2.1-alpha
scandone
6585 GWT:TIN:CONNECTING...
7087 GWT:TIN:CONNECTING...
7589 GWT:TIN:CONNECTING...
8091 GWT:TIN:CONNECTING...
scandone
state: 0 -> 2 (b0)
9301 GWT:TIN:CONNECTING...
state: 2 -> 3 (0)
state: 3 -> 5 (10)
add 0
aid 4
cnt 

connected with Mr-IoT.com, channel 1
dhcp client start...
ip:192.168.1.215,mask:255.255.255.0,gw:192.168.1.1
9809 GWT:TIN:CONNECTING...
9811 GWT:TIN:IP=192.168.1.215
9814 MCO:BGN:STP
9816 MCO:REG:NOT NEEDED
9818 MCO:BGN:INIT OK,TSP=NA
pm open,type:2 0
19755 GWT:TSA:C=0,CONNECTED
```

TCP connection:
```
$ nc 192.168.1.215 5003
0;255;3;0;14;Gateway startup complete.
[...]

PR #1064, opened 2018-02-18
ESP8266 Gateway examples: clarify inclusion pin

mfalkvidd:
I don't understand why README.md is modified, and I can't figure out how to reset it (tried git checkout -- README.md and git reset origin/development README.md but that made no difference).

Anyway,
    The comment for inclusion mode pin and the definition of
    MY_INCLUSION_MODE_BUTTON_PIN did not match.

    GPIO5 should be safe to use, while GPIO3 is used for the
    UART which might result in conflict.

    This fixes
    https://github.com/mysensors/MySensors/issues/993


fallberg:
It looks like you have not rebased your branch prior to filing the PR. The PR carry 5 commits (including the latest release commit).

mfalkvidd:
I have created a new branch. Will close this PR in favor of the new one.

PR #1067, opened 2018-02-18
Add git commit message hook to the mystools bundle

### LINKED ISSUE #672: Develop commit-msg hook to catch "bad commit messages" before git commit
Currently git commit messages are validated by Jenkins to ensure compliance with the MySensors' standards, a derivative of [Chris Beam's Seven Rules to writing good commit messages](http://chris.beams.io/posts/git-commit/).  The scope of this change request is to develop a client-side git commit-msg hook that shares the git commit message validation logic implemented on Jenkins.

This has a dependency on Jenkins CI scripts being refactored and added to the git repo.

/cc @fallberg

fallberg:
The message hook validates that the c ommit message follows
the MySensors requirements.
More specifically, it checks that
- Commit message does not exceed a line length of 72 characters
- Commit title start with an uppercase character
- Commit title does not end with a period ('.')

Fixes #672

### DIFF
--- a/.mystools/hooks/commit-msg.sh
+++ b/.mystools/hooks/commit-msg.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+too_long_lines=$(awk 'length>72' $1)
+if [[ ! -z $too_long_lines ]]; then
+	echo "Too long commit message lines:" $too_long_lines
+	has_failed=1
+fi
+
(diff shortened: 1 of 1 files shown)


fallberg:
At some point we might consider merging this logic with .ci/gitler.sh but these rules are pretty fixed and are not likely to change so it should suffice for now and provide developers with the means of having their commits pre-tested before pushing (or in this case even committing) code that is doomed to fail the toll-gate.

PR #1076, opened 2018-03-04
Fix default IRQ pins on ESP8266 that prevent MCU from boot

### LINKED ISSUE #1075: RFM69 on ESP8266 defaults to wrong pin
In both the old and the new driver, the IRQ pin for the RFM69 on ESP8266 defaults to GPIO2, which [equals to D4 on the Wemos D1 Mini](https://wiki.wemos.cc/products:d1:d1_mini#pin):
```
#elif defined(ARDUINO_ARCH_ESP8266)
#define DEFAULT_RFM69_IRQ_PIN                   (2)                                                                                             //!< DEFAULT_RFM69_IRQ_PIN
#define DEFAULT_RFM69_IRQ_NUM                   digitalPinToInterrupt(DEFAULT_RFM69_IRQ_PIN)    //!< DEFAULT_RFM69_IRQ_NUM
```
This 
1. is inconsistent with the [MySensors documentation](https://www.mysensors.org/build/connect_radio) which lists D1 as the interrupt port
2. prevents the ESP8266 from booting, because DIO0 is default low, but GPIO2 needs to be high for the ESP to boot

Because of 2. I cannot really imagine anyone is using it like that, so changing the default should not break anyones setup.

Also, I do not understand why for all platforms, `DEFAULT_RFM69_IRQ_NUM` defauls to `digitalPinToInterrupt(DEFAULT_RFM69_IRQ_PIN)` instead of `digitalPinToInterrupt(MY_RFM69_IRQ_PIN)`. The later would not require people to manually set both and avoid inconsistencies.

kvoit:
This PR fixes #1075 by changing the default IRQ pin for RFM69 and RFM95 on ESP8266 to D1, which is in accordance to the [documentation](https://www.mysensors.org/build/connect_radio#rfm6995-&-esp8266).

I also changed the default for IRQ_NUM in the old RFM69 driver to get the interrupt number from the configured, not the default pin. This is intentionally done in a separate commit.

See also [forum thread](https://forum.mysensors.org/topic/9083/cannot-get-rfm69hw-connection-between-esp8266gw-and-pro-mini-sensor/26)

### DIFF
--- a/drivers/RFM69/old/RFM69_old.h
+++ b/drivers/RFM69/old/RFM69_old.h
@@ -43,7 +43,7 @@
 #endif
 #define DEFAULT_RFM69_IRQ_NUM			digitalPinToInterrupt(MY_RFM69_IRQ_PIN)		//!< DEFAULT_RFM69_IRQ_NUM
 #elif defined(ARDUINO_ARCH_ESP8266)
-#define DEFAULT_RFM69_IRQ_PIN			(2)													//!< DEFAULT_RFM69_IRQ_PIN
+#define DEFAULT_RFM69_IRQ_PIN			(5)													//!< DEFAULT_RFM69_IRQ_PIN
 #define DEFAULT_RFM69_IRQ_NUM			digitalPinToInterrupt(MY_RFM69_IRQ_PIN)		//!< DEFAULT_RFM69_IRQ_NUM
 #elif defined(ARDUINO_ARCH_ESP32)
 #warning not implemented yet
--- a/drivers/RFM69/old/RFM69_old.h
+++ b/drivers/RFM69/old/RFM69_old.h
@@ -41,15 +41,15 @@
 #else
 #define DEFAULT_RFM69_IRQ_PIN			(2)													//!< DEFAULT_RFM69_IRQ_PIN
 #endif
-#define DEFAULT_RFM69_IRQ_NUM			digitalPinToInterrupt(DEFAULT_RFM69_IRQ_PIN)		//!< DEFAULT_RFM69_IRQ_NUM
+#define DEFAULT_RFM69_IRQ_NUM			digitalPinToInterrupt(MY_RFM69_IRQ_PIN)		//!< DEFAULT_RFM69_IRQ_NUM
 #elif defined(ARDUINO_ARCH_ESP8266)
 #define DEFAULT_RFM69_IRQ_PIN			(2)													//!< DEFAULT_RFM69_IRQ_PIN
-#define DEFAULT_RFM69_IRQ_NUM			digitalPinToInterrupt(DEFAULT_RFM69_IRQ_PIN)		//!< DEFAULT_RFM69_IRQ_NUM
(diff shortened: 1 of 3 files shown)


fallberg:
You can resolve the commit complaints by an interactive rebase and a force push (or squash the commits and amend the result).

drivers/RFM69/old/RFM69_old.h
```diff
@@ -41,15 +41,15 @@
 #else
 #define DEFAULT_RFM69_IRQ_PIN			(2)													//!< DEFAULT_RFM69_IRQ_PIN
 #endif
-#define DEFAULT_RFM69_IRQ_NUM			digitalPinToInterrupt(DEFAULT_RFM69_IRQ_PIN)		//!< DEFAULT_RFM69_IRQ_NUM
+#define DEFAULT_RFM69_IRQ_NUM			digitalPinToInterrupt(MY_RFM69_IRQ_PIN)		//!< DEFAULT_RFM69_IRQ_NUM
 #elif defined(ARDUINO_ARCH_ESP8266)
-#define DEFAULT_RFM69_IRQ_PIN			(2)													//!< DEFAULT_RFM69_IRQ_PIN
-#define DEFAULT_RFM69_IRQ_NUM			digitalPinToInterrupt(DEFAULT_RFM69_IRQ_PIN)		//!< DEFAULT_RFM69_IRQ_NUM
+#define DEFAULT_RFM69_IRQ_PIN			(5)													//!< DEFAULT_RFM69_IRQ_PIN
+#define DEFAULT_RFM69_IRQ_NUM			digitalPinToInterrupt(MY_RFM69_IRQ_PIN)		//!< DEFAULT_RFM69_IRQ_NUM
```
mfalkvidd: Since the behavior is the same for avr and esp8266, the define could be moved outside the #if defined clause. By having the same code in two places (like the current implementation), we risk forgetting to update one of the places in the future. The double code also makes the code harder to read and understand.

kvoit:
As I mentioned in the forum and he issue, I don' hink that the defaultts work with any ESP. GPIO2 needs to be pulled up for booing the sketch, and is actively pulled low by DIO0.
This is a property of the ESP8266, not the Wemos D1 Mini. The latter only provides the pullup resistors, that are overridden by DIO0.

fallberg:
@kvoit we cannot merge this PR until the toll gate issues have been resolved. If you have any questions, feel free to ask.

fallberg:
I still think the commit messages could be improved. Use a short title and be more elaborate in the body (separate title and body with a newline). This string gives very little info on what actually changed: "Get DEFAULT_RFM69_IRQ_NUM MY_..._PIN instead DEFAULT_..._PIN".
Also, please use imperative mode syntax in the title:
`Changed default IRQ pin for RFM69/95 on ESP8266 to GPIO5=D1` -> `Change default IRQ pin for RFM69/95 on ESP8266 to GPIO5=D1`

kvoit:
Done

PR #1078, opened 2018-03-06
Add credentials to violations plugin

fallberg:
The violations to github plugin was updated to require credentials.


fallberg:
Dropping PR. Setting can be applied globally in Jenkins.

PR #1093, opened 2018-03-25
Linux: Support keys in config file

marceloaqno:
- Signing (HMAC, serial) and encryption keys are set through the configuration file.
- SoftEeprom update.

### DIFF
--- a/configure
+++ b/configure
@@ -113,8 +113,20 @@ MySensors options:
 EOF
 }
 
+# Colors
+GREEN='\033[0;32m'
+RED='\033[0;31m'
+YELLOW='\033[1;33m'
+LIGHT_BLUE='\033[1;34m'
+NO_COLOR='\033[0m'
+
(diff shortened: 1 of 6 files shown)


fallberg:
This probably require the rPi specifics in MySigning.h doxygen block to be updated.

mfalkvidd:
Adding release-notes label since this change alters how personalization is done. instructions on mysensors.org will need to be updated.

fallberg:
I agree on the release notes but the homepage should not contain personalisation specifics. It should link to doxygen which will update automatically.

mfalkvidd:
You're right about the homepage. I confused these changes with https://github.com/mysensors/MySensors/pull/1061

fallberg:
Good idea with the label. Helps tracking those sneaky ones we might forget to mention on release day. Should reduce some forum chatter, or at least focus it a bit.

PR #1094, opened 2018-03-25
Linux: RF24 interrupt trigger workaround

### LINKED ISSUE #870: RPi: Stops receiving messages after some time when using nrf24 interrupts
This problem should have been fixed with #835, but my gateway still suffers from it although less frequent.
Related to #798.

@bathoryr Have you had any problems recently?

marceloaqno:
Fix #870

### DIFF
--- a/drivers/RF24/RF24.cpp
+++ b/drivers/RF24/RF24.cpp
@@ -410,11 +410,11 @@ LOCAL int16_t RF24_getSendingRSSI(void)
 LOCAL void RF24_irqHandler(void)
 {
 	if (RF24_receiveCallback) {
+#if defined(MY_GATEWAY_SERIAL) && !defined(__linux__)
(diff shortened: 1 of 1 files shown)


drivers/RF24/RF24.cpp
```diff
@@ -423,13 +423,28 @@ LOCAL void RF24_irqHandler(void)
 		// 1.Read payload, 2.Clear RX_DR IRQ, 3.Read FIFO_status, 4.Repeat when more data available.
 		// Datasheet (ch. 8.5) states, that the nRF de-asserts IRQ after reading STATUS.
 
+#if defined(__linux__)
```
mfalkvidd: Is there a reason we can't do the same for non-linux nodes?

marceloaqno: There is a small chance of losing packages, but since Linux is not a real-time system, we'll need a workaround

PR #1097, opened 2018-04-01
TinyGSM Gateway for MySensors

thucar:
TinyGSM adds support to a number of cheap GSM modems as well as the option to use an ESP8266 as a modem for an Arduino based gateway.

Example included.


thucar:
I give up. I can not seem to be able to please Jenkins/Gitler...

fallberg:
You need to do an interactive rebase and fix your commit messages. Adding new commits won't fix old commits.

fallberg:
I can also recommend installing the tools bundle. Instructions are on the contributing page on the MySensors homepage. That way, you can get coding style and commit messages validated when committing, which increases the chances of jenkins approving when pushing.

thucar:
I'm stumbling around somewhat blindly here, as I never got any e-mails from Jenkins describing what exactly seems to be the problem.

I'm using PHPStorm (old web development habbits die hard) as both, the editor and the version control software.

thucar:
I have no idea what Gitler/Jenkins could possibly be complaining about. He still refuses to send me any e-mails explaining his concerns.

mfalkvidd:
@thucar great work.
We'll need to add the new defines/macros to [MyConfig.h](https://github.com/mysensors/MySensors/blob/development/MyConfig.h#L1335) with Doxygen-style comments to make them appear in the [documentation](https://www.mysensors.org/apidocs-beta/index.html). That way, the new features are easy to find for users.

You are very welcome to do this yourself - it should be fairly easy, just follow the existing syntax in MyConfig.h. I'll be happy to help out if you need assistance or think it is too much work.

```
MY_GATEWAY_TINYGSM
MY_GSM_APN
MY_GSM_BAUDRATE
MY_GSM_PIN
MY_GSM_PSW
MY_GSM_RX
MY_GSM_SSID
MY_GSM_TX
MY_GSM_USR
```

fallberg:
Jenkins sends a mail to the email you use on github.

thucar:
I'm getting all other Gihub notifications, including when you reply here. However, nothing from Jenkins still.

mfalkvidd:
I found a message in Jenkins that seems to be related to sending email:
> Sending email to: ${env.CHANGE_AUTHOR_EMAIL}
> Error sending to the following INVALID addresses: ${env.CHANGE_AUTHOR_EMAIL}
@fallberg any idea why this happens?

mfalkvidd:
I have started on a new "Build" page for this gateway: https://www.mysensors.org/view/568

fallberg:
I suspect the authors local git config does not specify or specify an invalid email.

fallberg:
And I advice against merging to feature branches. Use rebase instead.

mfalkvidd:
@thucar does `git config --get user.email` on your local environment return something that looks like a valid email?

thucar:
@mfalkvidd Roger that, I'll get the defines/macros in the MyConfig.h file. For the general MY_GATEWAY_TINYGSM I had no problems finding a suitable spot, however with the more specific defines it's getting trickier. Shold I just create a section like this, somewhere suitable:

`#ifdef MY_GATEWAY_TINYGSM
/**
 * @def MY_GSM_APN
 * @brief Define MY_GSM_APN if you are using a GSM modem to connect to the Internet
 *
 * Description
 */
#define MY_GSM_BAUDRATE
/**
 * @def MY_GSM_APN
 * @brief Define MY_GSM_APN if you are using a GSM modem to connect to the Internet
 *
 * Description
 */
#define MY_GSM_BAUDRATE
/**
 * @def MY_GSM_PIN
 * @brief Define MY_GSM_APN if you are using a GSM modem to connect to the Internet
 *
 * Description
 */
#define MY_GSM_PIN
/**
 * @def MY_GSM_PSW
 * @brief Define MY_GSM_APN if you are using a GSM modem to connect to the Internet
 *
 * Description
 */
#define MY_GSM_PSW
/**
 * @def MY_GSM_RX
 * @brief Define MY_GSM_APN if you are using a GSM modem to connect to the Internet
 *
 * Description
 */
#define MY_GSM_RX
/**
 * @def MY_GSM_SSID
 * @brief Define MY_GSM_APN if you are using a GSM modem to connect to the Internet
 *
 * Description
 */
#define MY_GSM_SSID
/**
 * @def MY_GSM_USR
 * @brief Define MY_GSM_APN if you are using a GSM modem to connect to the Internet
 *
 * Description
 */
#define MY_GSM_USR
#endif`

And yup, `git config --get user.email` does return my correct email address

fallberg:
This is the patch jenkins try to send:
https://ci.mysensors.org/job/MySensors/job/MySensors/job/PR-1097/10/execution/node/3/ws/MySensors/restyling.patch/*view*/

thucar:
Thanks @fallberg! I'm on it

mfalkvidd:
@fallberg in email troubles happen again, would it be easy to att the link to the patch file in the gitler [report](https://ci.mysensors.org/job/MySensors/job/MySensors/job/PR-1097/10/Gitler_report/gitler.html)?

fallberg:
Perhaps the failed mailing could be caused by email privacy settings of your github account. AFAIK jenkins usually is able to mail committers.

fallberg:
@mfalkvidd it is tricky as the workspace is not persistent

mfalkvidd:
@thucar thanks for checking. Strange that the email address doesn't get through to Gitler.

For the more specific defines, just add them somewhere within the `#if doxygen` section. And don't include `#ifdef MY_GATEWAY_TINYGSM`. Doxygen is a special build mode where all features are turned on (so they show up in the documentation).

If you can, include an example of what the parameter value can be. Example:
```
@def MY_GSM_APN
@brief APN from your cell carrier / mobile provider. Example: 4g.tele2.se
```

mfalkvidd:
@fallberg could we get a more verbose output of the email call that fails? Preferably with the real value of  ${env.CHANGE_AUTHOR_EMAIL}, if possible.

mfalkvidd:
When the keywords have been added to MyConfig.h, Gitler will start complaining that they aren't included in keywords.txt (which provides syntax highlighting in the Arduino IDE). Just add them all wherever you find appropriate. And make sure there is a TAB between the keyword and LITERAL1. The Arduino IDE doesn't work if any other whitespace character is used.

fallberg:
@mfalkvidd I suspect the value is empty for some reason. I am only on phone so I am limited in what I can see, do and try.

fallberg:
@mfalkvidd if you log in to jenkins, you can replay a job and hack the groovy to print some more stuff

keywords.txt
```diff
@@ -103,6 +103,17 @@ MY_NODE_LOCK_COUNTER_MAX	LITERAL1
 MY_HW_HAS_GETENTROPY	LITERAL1
 hwGetentropy	KEYWORD2
 
+# TinyGSM
+MY_GATEWAY_TINYGSM  LITERAL1
```
mfalkvidd: The Arduino IDE won't tolerate any whitespace except a single TAB between the keyword and LITERAL1. This code has spaces.

thucar: This is very odd as I’m specifically using tab. Might be PHPStorm not playing nice. I’ll look into it.

mfalkvidd:
Arduino IDE 1.8.6 will **finally** be fixed to handle whitespace in keywords.txt (see https://github.com/arduino/Arduino/pull/6693 ) but since that version hasn't been released yet it would be nice if we could use tab.

mfalkvidd:
I think the include needs to look like this
` #include "drivers/Linux/EthernetClient.h"`
For Jenkins, I guess we'll need to exclude the driver code. I am not sure how to do that though. Something similar must have been done for the other drivers. Maybe @fallberg knows?

fallberg:
If there are examples that does not work for certain boards, yes, they need to be filtered in the groovy in the .ci folder.
See arduino.groovy.

fallberg:
Sketches that depend on third party drivers are placed in the arduino examples repo, not the core library repo.

thucar:
Ok, so the only option is to make a hard copy of the 3rd party library, place it in the drivers/ folder and just go over their code with a fine comb to make Jenkins happy? Gotcha, I'm on it.

fallberg:
No, your pr is filed against the wrong repo. There is an examples repo for sketches that depend on third party libraries.

fallberg:
https://github.com/mysensors/MySensorsArduinoExamples

fallberg:
Sorry I did not notice the external driver dependency earlier.

fallberg:
But your suggestion is also an option. But I think more core team members need to weigh in if we want to add more driver copies to the core repo. My personal opinion is that we should avoid drivers the core itself does not depend on. The other repo was created for this exact purpose.

thucar:
Ok, I'm at a loss here. 

Couple questions:
1) The example is not the one to include the conflicting TinyGSM library. The library gets included by MySensors.h Is that incorrect? Should external libraries be left to include by the actual sketch instead of trying to include them when we detect they are needed (#ifdef MY_GATEWAY_TINYGSM)
2) Should I not include my example in this PR, but create a separate PR against the examples repository just for the example?

thucar:
Ok, you were too quick for me :) I've included the library from the sketch itself and removed calls to it from MySensors.h Let's see if Jenkins approves

fallberg:
I think not. Jenkins does not pull external drivers. It is designed to validate the core library against all supported boards. Only a limited set of examples are included which does not depend on external libraries.

fallberg:
Since you added a new example it will be picked up by jenkins but it will not be able to validate it if 8t depend on third party libraries. So it comes down to either moving the example to the examples repo or include the external driver in the core repo (here). I have already stated my preference :) but your contribution is highly appreciated and equally valuable if moved to the examples repo. Most examples already live there. But it is a big task to maintain jenkins to support all kinds of obscure external libraries that all those examples depend on.

thucar:
I'm failing to understand how Jenkins is able to verify the code I've added to the core, if that code references an external library :D

fallberg:
That's my point. It cannot. So you either need to file your or pr against the other repo, or include the driver in this repo. Jenkins does not care about the other repo

fallberg:
But you also need to make sure you don't enable features in the core that make it fail existing examples.

fallberg:
For example, including external drivers by default and not have them feature toggled. Existing examples in the core should work before and after you have made changes to the core. Jenkins is there to make sure changes does not break existing examples on the core.

thucar:
@fallberg The code in question resides mainly in the core/MyGatewayTransportMQTTClient.cpp file

There are no core files on the other repo. I do agree with you that including non-core dependency libraries in the drivers/ folder does not make sense. But at the moment, I see no other option?

fallberg:
I see that. But examples fail, so you have something that by default include unknown stuff (or it fails for some other reason, the details link points you to the problem).

fallberg:
I am on phone so browsing code is a pain, but I can see that your change set is core specific so yes, it is targeting the correct repo. I'll try to see what jenkins does not like.

thucar:
Ah, ok. Now I'm starting to get it. In my last amend I removed the example file. And without the example file there should be nothing to toggle the changes I've added.  So hopefully, Jenkins will let me pass this time :)

fallberg:
Yep, that's probably it.

fallberg:
You can then add your example to the other repo and test it locally when this pr is submitted.

fallberg:
(after pulling the development branch)

mfalkvidd:
To me, the TinyGSM library is analogous to the Ethernet library or the PubSubClient library. Examples for Ethernet and MQTT gateways are part of the MySensors github repository, they are not in the MySensorsArduinoExamples github repository.

But I guess it has been quite a while since we added a new dependency in the MySensors github repository, so we have no rules or process for how to do it. Maybe we should pull more core developers into this to get a discussion?

thucar:
Well, at least I can finally get along with Jenkins!

fallberg:
Good work! 
And yes, I would also like more voices regarding the driver. We now have features in the core that has dependency to a driver, which makes the driver eligible for inclusion in the core. And on turn, an example that "exercise" jenkins with the added feature and driver. Since we want all core features to be verifiable by jenkins (and have users use core features without messing with external dependencies). 
So perhaps either this pr gets more commits that add the driver and an example for it. Jenkins is happy with the current state. Or amend the existing commit. Either way, the commits will be squashed.
Or just file new pr(s) for the driver and example later on.

thucar:
Cool, I'll get to cleaning up the TinyGSM code. I'll squash the commits into one in the end.

fallberg:
You can relax jenkins on the doxygen requirements and warnings requirements by adding to the excludePattern in doxygen.groovy in  .ci. Similar works for the warnings module in the other groovies.

fallberg:
For the added third party driver I mean. You should find similar things for some other drivers.

mfalkvidd:
I'd say do more work in new PRs later. This PR is in itself a great addition to MySensors. No neet to make it more complicated now.

thucar:
@mfalkvidd Too late, I think I got this :+1:

fallberg:
Cppcheck issues you do not agree with can be in-line suppressed. There are a few occasions around in the repo to use as reference.

thucar:
Thank you for the guidance @fallberg :+1:  I'm getting there. Slowly, but surely.

fallberg:
A heads up on the examples toll-gates. You might need to filter your new example for some boards. The gateways have trouble fitting the cheaper targets. This is ok, but jenkins need to be told not to attempt to build certain examples for certain boats. arduino.groovy contain some examples of how to handle this. Thank you for your patience.

drivers/TinyGSM/TinyGsmFifo.h
```diff
@@ -0,0 +1,153 @@
...
+#define TinyGsmFifo_h
+template <class T, unsigned N>
+class TinyGsmFifo
+{
+public:
+	// Suppress warning about uninitialized variables because initializing them in an init function
+	// allows the compiler to optimize away the variables in case the class is only instantiated but
+	// never used.
+	// cppcheck-suppress uninitMemberVar
+	TinyGsmFifo()
```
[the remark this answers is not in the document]
thucar: Come on man, the suppress message is right there. Read it!

thucar:
Could someone offer a bit of advice on this one? Either how do I initialize the member variable in the construct as Jenkins wishes, or how to properly use the suppress. I'm at my wits end here...

drivers/TinyGSM/TinyGsmFifo.h
```diff
@@ -0,0 +1,153 @@
+#ifndef TinyGsmFifo_h
+#define TinyGsmFifo_h
+template <class T, unsigned N>
+class TinyGsmFifo
```
fallberg: I believe the suppression need to be placed here, just before the class declaration.

fallberg: See here for reference: https://github.com/mysensors/MySensors/blob/b5a2fc226fe64759f2cb47a255b9b6f34850a968/drivers/ATSHA204/sha256.cpp#L36

drivers/TinyGSM/TinyGsmFifo.h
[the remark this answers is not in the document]
fallberg: It could also be that the suppression has to be on the line the warning is issued on, which is 11, not 10.
That is,
`TinyGsmFifo() // cppcheck-suppress uninitMemberVar`

fallberg:
It could be that the suppression has to be on the line the warning is issued on, which is 11, not 10.
That is,
`TinyGsmFifo() // cppcheck-suppress uninitMemberVar`
or the suppression needs to be made just before the class declaration (which seem to be the case on similar situations.

drivers/TinyGSM/TinyGsmFifo.h
```diff
@@ -0,0 +1,152 @@
...
+		return n - c;
+	}
+
+private:
+	int _inc(int i, int n = 1)
+	{
+		return (i + n) % N;
+	}
+
+	T    _b[N];
```
fallberg: I am not sure why it refuse to acknowledge your suppression. Perhaps it is needed here?

fallberg: I am not sure why it refuse to acknowledge your suppression. Perhaps it is needed here?

fallberg:
I suggest we take a different approach since the inline cppcheck suppression for this particular issue is non-intuitive (and therefore runs the risk of being deleted at some point), you update https://github.com/mysensors/MySensors/blob/development/.mystools/cppcheck/config/suppressions.cfg and add this:
`uninitMemberVar:*/TinyGsmFifo.h`

thucar:
@fallberg *highfive*

.ci/arduino.groovy
```diff
@@ -206,6 +211,7 @@ def buildEsp8266(config, sketches, String key) {
 					sketches[sketch].path != config.library_root+'examples/GatewaySerialRS485/GatewaySerialRS485.ino' &&
 					sketches[sketch].path != config.library_root+'examples/GatewayW5100/GatewayW5100.ino' &&
 					sketches[sketch].path != config.library_root+'examples/GatewayW5100MQTTClient/GatewayW5100MQTTClient.ino' &&
+					sketches[sketch].path != config.library_root+'examples/GatewayGSMMQTTClient/GatewayGSMMQTTClient.ino' &&
```
fallberg: Is your GSM gateway not supposed to be tested with at least one of the supported boards? In theory, I suppose it will work on any board with enough memory?

thucar: I have no idea which hardware would be able to run it and which would not, so I just tried to follow the lead set by other MQTT gateway examples. However, looking at it right now it seems I followed the lead of ESP8266 example. Whoops.

I went back and only left tests disabled for the same hardware that was not allowing W5100MQTTClient examples.

fallberg: That sounds reasonable.

thucar:
I think I caused Jenkins to have a nervous breakdown...

fallberg:
Hm, looks like something changed causing that particular warning to disappear, which would cause the suppression to throw that message. So it looks like you can revert your change to suppressions.cfg.

drivers/TinyGSM/TinyGsmClient.h
```diff
@@ -0,0 +1,67 @@
...
+typedef TinyGsm::GsmClient TinyGsmClient;
+
+#elif defined(TINY_GSM_MODEM_M590)
+#define TINY_GSM_MODEM_HAS_GPRS
+#include "TinyGsmClientM590.h"
+typedef TinyGsm::GsmClient TinyGsmClient;
+
+#elif defined(TINY_GSM_MODEM_U201)
+#define TINY_GSM_MODEM_HAS_GPRS
+#include "TinyGsmClientU201.h
```
fallberg: I think this line miss a " :)

thucar: You are right! I bet it was that Jenkins character who swiped it. He seems dodgy to me, I don't trust him....

drivers/TinyGSM/TinyGsmFifo.h
```diff
@@ -0,0 +1,152 @@
+#ifndef TinyGsmFifo_h
+#define TinyGsmFifo_h
+
+template <class T, unsigned N>
+
+class TinyGsmFifo
+{
+public:
+
+	TinyGsmFifo()
```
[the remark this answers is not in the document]
thucar: Enabling the suppression again. Probably that missing " caused the previous condition where it got registered as a false positive?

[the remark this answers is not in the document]
fallberg: Yes, could be that it threw the syntax parser off completely.

thucar:
I'm switching to SoftwareSerial as the default option for my example to have it pass all hardware checks. As a future upgrade I'll add a check in the MySensors.h to draw users attention to switch to Soft serial when they are using hardware that does not have hardware serial.

thucar:
Hmm.. SoftwareSerial is not available for Jenkins?

fallberg:
It depends on the target. The MySensors gw is on samd architecture. Now you see the power of jenkins :) it is needed to make sure our core remains in a compilable state across a multitude of architectures.

thucar:
I’ve already come to appreciate Jenkins a great deal. I’ve learned more about git and different Arduino plarforms in the last couple days than I did in the last couple years. :)

@mfalkvidd hopefully you do not mind me taking a bit longer than initially planned. I’d rather get this thing done properly.

mfalkvidd:
Not at all :)
While you're at it, how about upstreaming the TinyGSM changes to their source? ;-)

mfalkvidd:
About SoftwareSerial; my guess is that most users would want to use software serial for GSM communication so they can use the hardware serial for MySensors debug output. Troubleshooting blindly is no fun.

fallberg:
SAMD devices has a couple of serdes devices. @tbowmo did the MySensors port, he perhaps knows how to best handle it for the MySensorsGW.
Regarding other boards, the major ports are "arduino native", SAMD, nRF5 and ESP. There is also Linux/rPi but I guess that needs more testing before we just "blindly" offer the support for that target unless you have something to test with.

mfalkvidd:
I don't think this solution for GSM backhaul would make sense on the raspberry pi gateway. If someone wants GSM backhaul on Raspberry Pi there are many solutions that are easier to use - a gsm usb dongle for example.

fallberg:
True, but it's always nice with options :) But I agree completly, the effort is probably not worth the benefit.

thucar:
Looks like I got it all sorted. Finally :)

MySensors.h
```diff
@@ -153,7 +153,7 @@ MY_DEFAULT_RX_LED_PIN in your sketch instead to enable LEDs
 
 // GATEWAY - COMMON FUNCTIONS
 // We support MQTT Client using W5100, ESP8266 and Linux
```
fallberg: Perhaps update this comment to also include GSM as communications medium now

core/MyGatewayTransportMQTTClient.cpp
```diff
@@ -50,8 +50,24 @@ IPAddress _subnetIp(255, 255, 255, 0);
 uint8_t _MQTT_clientMAC[] = { MY_MAC_ADDRESS };
 #endif /* End of MY_GATEWAY_ESP8266 */
 
+#if defined(MY_GATEWAY_TINYGSM)
```
fallberg: We should probably indicate that the GSM gateway only applies to MQTT GW:s either though the define or at least in the doxygen docs for the flag. If users enable the MY_GATEWAY_TINYGSM flag, and don't indicate MQTT it won't work.
The example does this, but in case people forget that flag when rolling their own gw it might cause some forum "noise".
An option could be to put a guard in MyConfig.h, along the lines of this:
```
#if defined(MY_GATEWAY_TINYGSM) && !defined(MY_GATEWAY_MQTT_CLIENT)
#error MY_GATEWAY_TINYGSM only works with MY_GATEWAY_MQTT_CLIENT
#endif
```

MySensors.h
thucar: Done!

core/MyGatewayTransportMQTTClient.cpp
thucar: Good idea. I also added a check to make sure both TX and RX pins are defined when using softserial

MySensors.h
```diff
@@ -307,6 +316,11 @@ MY_DEFAULT_RX_LED_PIN in your sketch instead to enable LEDs
 #include "drivers/AVR/DigitalIO/DigitalIO.h"
 #endif
 
+// SOFTSERIAL
+#if defined(MY_GSM_TX) != defined(MY_GSM_RX)
```
fallberg: Is it intentional that the error will fire if both flags are defined or both are not defined?

thucar: hrm, I guess I'm missing the logic behind the statement. The goal was to fire an error if both are not the same (i.e. one is defined and the other is not)

thucar: And testing it right now, it would seem to behave as I had hoped.

fallberg: Then I think what you want is something like
´#if (defined(MY_GSM_TX) != defined(MY_GSM_RX)) && defined(MY_GSM_TX)´

fallberg: Ok, well if it works as you like it, it looks fine by me. But I would assume that if you don't define either of them, and enter this segment, the error will fire.

thucar: What is the benefit of the added ´&& defined(MY_GSM_TX)´ ?

I just ran the tests with all three conditions: both enabled, one enabled and both disabled, and the error message behaved as it should. I did however find that I need to add a separate check to make sure if softserial pins are not defined, then the hw serial must be available.

fallberg: I might be misunderstanding the conditions for the definitions,. but I would assume that it evaluates the expression in boolean manner, so the two defined() blocks either return true or false depending on the definition state, and the != comparison will evaluate to true if either of the states is different from the other. No matter if they are true or false.
The result should then be that the expression evaluates to true given that both define() blocks return true or false.

fallberg: I see now, the error of course kicks in if only one is set. If none or both are set, it is ok. My bad.

thucar: All good :) Your comment made me run more extensive tests and discover the parallel problem of errors when not defining any Serial port for the modem to use. Got that sorted as well now :+1:

fallberg: Excellent.

fallberg:
Big thanks for the contribution!

Issue #1096, opened 2018-03-26
Update GatewayESP8266 and GatewayESP8266MQTTClient wiring


mfalkvidd:
The wiring instructions provided in comments in GatewayESP8266 and GatewayESP8266MQTTClient  example sketches do not match the wiring instructions at https://www.mysensors.org/build/connect_radio

Suggestion: Remove the wiring instructions entirely from both sketches and point to https://www.mysensors.org/build/connect_radio only.

This discrepancy was reported by forum user RWoerz in https://forum.mysensors.org/topic/4800/building-a-wifi-gateway-using-esp8266/58

thucar:
I believe the wiring instructions in those examples should be carried over to the "Connecting the radio" page. The instructions are correct, but for a bare ESP8266 module whereas the current instructions on the Radio page are for the NodeMCU/Wemos boards.

I'm using mostly ESP8266-12E's on their own and always found myself glancing at the wiring instructions in the examples to refresh my memory.

mfalkvidd:
Thanks for the feedback on ESP-12E. I am not familiar with it, so I just left the comments in the sketch.
Now I've googled ESP-12E and I realized that I bought a few some time ago, but never got around to use them since the Wemos D1 mini is so much easier to use :-)

Just to make sure we're talking about the same instructions, these are the instructions **I** believe  we're discussing:
```
 * Not all ESP8266 modules have all pins available on their external interface.
 * This code has been tested on an ESP-12 module.
 * The ESP8266 requires a certain pin configuration to download code, and another one to run code:
 * - Connect REST (reset) via 10K pullup resistor to VCC, and via switch to GND ('reset switch')
 * - Connect GPIO15 via 10K pulldown resistor to GND
 * - Connect CH_PD via 10K resistor to VCC
 * - Connect GPIO2 via 10K resistor to VCC
 * - Connect GPIO0 via 10K resistor to VCC, and via switch to GND ('bootload switch')
```

To me, adding these instructions to "connecting the radio" feels a bit weird, since none of the instructions concern the radio(s).

Would a new section at https://www.mysensors.org/build/esp8266_gateway be a good place?

PR #1100, opened 2018-04-02
Remove faulty wiring instructions from ESP gateways

mfalkvidd:
This fixes
https://github.com/mysensors/MySensors/issues/1096

Thanks to RWoerz for reporting this discrepancy.

### DIFF
--- a/examples/GatewayESP8266/GatewayESP8266.ino
+++ b/examples/GatewayESP8266/GatewayESP8266.ino
@@ -30,34 +30,20 @@
  *
  * VERA CONFIGURATION:
  * Enter "ip-number:port" in the ip-field of the Arduino GW device. This will temporarily override any serial configuration for the Vera plugin.
- * E.g. If you want to use the defualt values in this sketch enter: 192.168.178.66:5003
+ * E.g. If you want to use the default values in this sketch enter: 192.168.178.66:5003
  *
  * LED purposes:
  * - To use the feature, uncomment any of the MY_DEFAULT_xx_LED_PINs in your sketch, only the LEDs that is defined is used.
- * - RX (green) - blink fast on radio message recieved. In inclusion mode will blink fast only on presentation recieved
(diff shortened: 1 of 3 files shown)


thucar:
I'm referring to the entire wiring section below:
```
 * See http://www.mysensors.org/build/esp8266_gateway for wiring instructions.
 * nRF24L01+  ESP8266
 * VCC        VCC
 * CE         GPIO4
 * CSN/CS     GPIO15
 * SCK        GPIO14
 * MISO       GPIO12
 * MOSI       GPIO13
 * GND        GND
 *
 * Not all ESP8266 modules have all pins available on their external interface.
 * This code has been tested on an ESP-12 module.
 * The ESP8266 requires a certain pin configuration to download code, and another one to run code:
 * - Connect REST (reset) via 10K pullup resistor to VCC, and via switch to GND ('reset switch')
 * - Connect GPIO15 via 10K pulldown resistor to GND
 * - Connect CH_PD via 10K resistor to VCC
 * - Connect GPIO2 via 10K resistor to VCC
 * - Connect GPIO0 via 10K resistor to VCC, and via switch to GND ('bootload switch')
```

The link to http://www.mysensors.org/build/esp8266_gateway currently does not actually provide any wiring or operating instructions for the use of barebone ESP8266 modules. 

I agree that if the above information could be placed on that page, there would be no more need for it in the examples nor the radio page.

mfalkvidd:
@thucar what board settings do you use in the Arduino IDE for ESP-12E?
I am trying to figure out if the _GPIOn_ <-> _Dn_ mappings for the radio pins result in the same radio wiring as described on https://www.mysensors.org/build/connect_radio

If they are correct, I suggest we
1. add the GPIO names to the wiring table on https://www.mysensors.org/build/connect_radio (so both Dn and GPIOn are available).
2. add the boot wiring to the ESP8266 page (since the boot wiring is not related to the radio wiring).

PR #1106, opened 2018-04-07
Clarify default defines

mfalkvidd:
Moved example values to the @ brief portion of Doxygen,
so the default (empty/undefined) values are visible
in Doxygen.


mfalkvidd:
Before this change:
image
after this change:
image

mfalkvidd:
@fallberg how about this? Brief is kept short and entire example is in monospace.
image

PR #1108, opened 2018-04-08
Add keywords that already have Doxygen

mfalkvidd:
Found some keywords that already have Doxygen comments
but were not present in keywords.txt


keywords.txt
```diff
@@ -282,10 +288,14 @@ MY_ESP8266_SSID	LITERAL1
 # MY_DEBUG_VERBOSE_CORE
 # MY_GATEWAY_FEATURE
 # MY_RAM_ROUTING_TABLE_ENABLED
+# MY_RF24_CONFIGURATION
```
tekka007: > internal define

keywords.txt
```diff
@@ -282,10 +288,14 @@ MY_ESP8266_SSID	LITERAL1
...
+# MY_RFM69HW
```
tekka007: internal define

mfalkvidd: @tekka007 did you see the comment at top of this block? https://github.com/mfalkvidd/MySensors/blob/67bccd15b8feda9c94a3267b968ad5e32bff68f1/keywords.txt#L278

Issue #1109, opened 2018-04-08
Consolidate wifi defines


mfalkvidd:
MySensors dev currently uses the following defines for wifi:
```
MY_ESP32_BSSID
MY_ESP32_HOSTNAME
MY_ESP32_PASSWORD
MY_ESP32_SSID
MY_ESP8266_BSSID
MY_ESP8266_HOSTNAME
MY_ESP8266_PASSWORD
MY_ESP8266_SSID
```
The board is not very relevant, and by consolidating the defines we would only need to document them once. They could look like this:
```
MY_BSSID
MY_HOSTNAME
MY_PASSWORD
MY_SSID
```
- [x] Add backwards compatibility to support MY_ESP8266_* defines in existing sketches (ifdef MY_ESP8266_SSID "xyz" #define MY_SSID "xyz")
- [x] Move MY_ESP8266_* keywords to the deprecated section in keywords.txt so they are no longer highlighted
- [x] Change ESP32 code to only use the new defines. No backwards compatibility needed since ESP32 support is brand new
- [x] Add Doxygen comments to the new defines

marceloaqno:
I think MY_PASSWORD could cause some confusion with signing simple password feature.
How about this?
MY_HOSTNAME
MY_WIFI_BSSID
MY_WIFI_PASSWORD
MY_WIFI_SSID

thucar:
Would it make sense to also use the above for MY_GATEWAY_TINYGSM when using an Esp8266 as a modem?

mfalkvidd:
Esp8266 defines updated by https://github.com/mysensors/MySensors/pull/1110

PR #1110, opened 2018-04-09
Consolidate WiFi defines

tekka007:
see #1109


examples/GatewayESP32/GatewayESP32.ino
```diff
@@ -38,12 +38,12 @@
 
 #define MY_GATEWAY_ESP32
 
-#define MY_ESP32_SSID "MySSID"
-#define MY_ESP32_PASSWORD "MyVerySecretPassword"
+#define WIFI "MySSID"
```
mfalkvidd: Should be MY_WIFI_SSID?

keywords.txt
```diff
@@ -345,6 +341,14 @@ MY_ESP8266_SSID	LITERAL1
 # MY_RF69_RESET
 # MY_RF69_SPI_CS
 # MY_TRANSPORT_DONT_CARE_MODE
+# MY_ESP8266_BSSID	LITERAL1
+# MY_ESP8266_HOSTNAME	LITERAL1
+# MY_ESP8266_PASSWORD	LITERAL1
+# MY_ESP8266_SSID	LITERAL1
+# MY_ESP32_SSID	LITERAL1
```
mfalkvidd: Maybe we don't need to blacklist the esp32 keywords since they were never released?

core/MyGatewayTransportEthernet.cpp
```diff
@@ -120,36 +149,31 @@ bool gatewayTransportInit(void)
 #if defined(MY_ESP8266_SSID)
```
mfalkvidd: For this section, it would be possible to consolidate the ESP8266 and ESP32 code. The only difference seems to be WiFi.hostname for ESP8266 vs Wifi.setHostname for ESP32.

Would it make sense to do
```
#if defined MY_GATEWAY_ESP8266
WiFi.hostname(MY_HOSTNAME)
#else if defined MY_GATEWAY_ESP32
WiFi.setHostname(MY_HOSTNAME)
#endif
```
and use the same code for everything else?

PR #1113, opened 2018-04-19
Add sleep forever by sleep(0);

Tombula:
For battery powered sensors which wake up by the RST Pin its useful to sleep the cpu forever. When using this command: sleep(INTERRUPT_NOT_DEFINED, MODE_NOT_DEFINED, 0, false); 
the code will be stripped down to hwSleep(0);. But this will not sleep the cpu forever.

### DIFF
--- a/hal/architecture/AVR/MyHwAVR.cpp
+++ b/hal/architecture/AVR/MyHwAVR.cpp
@@ -141,8 +141,23 @@ void hwInternalSleep(uint32_t ms)
 
 int8_t hwSleep(uint32_t ms)
 {
-	hwInternalSleep(ms);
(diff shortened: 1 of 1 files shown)


mfalkvidd:
Looks like this problem was introduced by https://github.com/mysensors/MySensors/pull/814

mfalkvidd:
@Tombula would you mind amending the commit message to make Gitler happy? Just run git commit --amend and manually add newlines to the commit message. Then force push your branch.

Tombula:
@mfalkvidd I'm very new to git and it's commit system. I just did the commit using the website github.com. But I think you want me to clone the Repository on my system, change the specific file and than execute git commit --amend?
Thank you very much for your time :)

PR #1124, opened 2018-05-01
Kill Gitler and make way for The Butler

fallberg:
The Butler is a more polite and helpful replacement to the grumpy old Gitler.
The Butler will do his utmost to assist new (and old) contributors to the MySensors cause.


.ci/butler.groovy
```diff
@@ -115,31 +115,35 @@ def call(config) {
...
-							./gitler.sh""")
+							./butler.sh""")
 
 	if (fileExists(config.repository_root+'restyling.patch')) {
 		emailext (
-			subject: "Job '${env.JOB_NAME} #${env.BUILD_NUMBER} [PR#${env.CHANGE_ID}]' failed due to bad code styling",
-			body: """<p>Job '${env.JOB_NAME} [<a href="${env.CHANGE_URL}">PR#${env.CHANGE_ID}</a> - ${env.CHANGE_TITLE}]' failed because code style does not follow the standards.</p>
-			A patch to rectify the errors is attached. You apply the patch using:<br>
+			subject: "PR#${env.CHANGE_ID} - ${env.CHANGE_TITLE} has unfortunate code styling",
+			body: """<p>Greetings Sir!<p>
```
MartinHjelmare: This assumes the person submitting the PR is a man. Shouldn't we try to be as inclusive as possible?

MartinHjelmare: Would it hurt the text to remove the use of "Sir" by the butler?

mfalkvidd: The butler uses Master in "Well done Master" which isn't very gender-neutral either. Would be nice to use the same word everywhere.
Maybe use "Highness" or "Your Grace"?

We want the butler to be perceived as helpful and polite (instead of harsh and demanding) so some sort of polite expression would be nice.

fallberg: Well, Well done Master is exclusive for core team members, and last time I checked, it was not very gender-neutral either.
If Sir is not good enough, please suggest alternatives. "Highness" and "Your Grace" sounds a bit odd and overly arcane I think.

MartinHjelmare: I would remove all titles altogether.

fallberg: Ok, lets go for that. I see the PR is already merged. Will you file a new one?

MartinHjelmare: Ok, I'll take a shot.

PR #1131, opened 2018-05-14
Add an error message about missing abstraction layer

Avamander:
This would let the user of the library know the platform they're using is not supported instead of throwing out hundreds of errors.

### DIFF
--- a/MySensors.h
+++ b/MySensors.h
@@ -77,6 +77,8 @@
 #include "hal/architecture/Teensy3/MyHwTeensy3.cpp"
 #elif defined(__linux__)
 #include "hal/architecture/Linux/MyHwLinuxGeneric.cpp"
+#else
+#error Hardware abstraction not defined (unsupported platform)
 #endif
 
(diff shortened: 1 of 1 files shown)


fallberg:
I am not sure why Cppcheck fails on this commit. Could you please make sure you have rebased your branch against the latest development upstream?

Avamander:
I used GitHub's web interface, I don't think it allows using anything other than the latest upstream?

fallberg:
Hm, not sure if the web interface does anything strange then. Other PRs don't get this error.

fallberg:
I am not a fan of the commit message. It does not provide useful information on the change. "update mysensors.h" can mean anything.

Avamander:
Hmm, fair enough, squash merge allows to change that though?

fallberg:
Absolutely. Suggestion for title/message?

Avamander:
Maybe "Added an error message when HAL couldn't be determined"?

fallberg:
Thanks for the contribution.

PR #1137, opened 2018-05-27
avoid data array overflow

seeers:
can caused by package collisions


fallberg:
Changing the title of a PR does not change the commit headline. Instructions are provided in the Butler-report.

PR #1154, opened 2018-06-18
Development version 2.3.1-alpha

fallberg:


core/Version.h
```diff
@@ -50,7 +50,7 @@
 #define MYSENSORS_LIBRARY_VERSION_MINOR							3							//!< Minor release version
 #define MYSENSORS_LIBRARY_VERSION_PATCH							0							//!< Patch version
```
tekka007: should be 1 for 2.3.1-alpha

fallberg: Indeed. Automation fail.

fallberg: Requested change done

PR #1170, opened 2018-08-04
ESP32: Update to support 1.0.0 board defs

tekka007:
- WIP: ESP32 board defs added using Arduino IDE board manager => path problems using arduino-builder
- ESP32 nightly builds fetched via cron job


tekka007:
implemented in #1244

PR #1175, opened 2018-08-10
Added support for pre configured HC12 serial Radio

tipih:
This is first version for support of the HC12 serial radio module. For the time being the Radio need to be preconfigured in pair.

- **Baudrate** 9600 when using software serial, default value in the sw is 9600
- **Channel** between 1-127 is done with AT+Cxxx
- **Power** between 1-8, Default value is 8 +20dBm to high when when working with the radio on desk, suggest to change to 1 = -1dBm I had issues when using full power, might also be the LDO.

I have the following in the config file
MY_HC12_BAUD_RATE
MY_HC12_MAX_MESSAGE_LENGTH
MY_HC12_SOH_COUNT
MY_HC12_SET_PIN
MY_HC12_TXPIN
MY_HC12_RXPIN
MY_HC12_PA_LEVEL

primary function is in 
MyTransportHC12, that is based on the MyTransportRS485 code.

The radio works fine on ESP8266, Arduino Nano, pro mini and Uno, that is what I have currently tested
Currently it does not compile for Attiny, but that might be releated to the Software serial used, insted of AltSoftwareSerial.

I do not know what to change in the file called configure, not sure what it does.

I will be adding full support some of the settings, like Baud_Rate, Channel Power and FU1-3, I will also see if I get get sleep to work.
However in FU2 Idle current is only 80uA, that should be ok for battery driven sensors.

The HC12 works on a very long distance, and is fairly cheep.


tipih:
Help needed.
I got the toolchain installed and it seams to work, but I get this error, btw. the code works very nicely, and the HC12 radio is doing it job fine and stable.


michael@michael-PC MINGW64 ~/Documents/Arduino/libraries/MySensors/hal/transport/HC12 (hc12Radio)
$ git commit MyTransportHC12.cpp -m "Fix of style"
[drivers\Linux\Stream.h:52] -> [drivers\Linux\SerialPort.h:73]: (style) The function 'available' overrides a function in a base class but is not marked with a 'override' specifier.
[drivers\Linux\Stream.h:53] -> [drivers\Linux\SerialPort.h:79]: (style) The function 'read' overrides a function in a base class but is not marked with a 'override' specifier.
[drivers\Linux\Print.h:60] -> [drivers\Linux\SerialPort.h:86]: (style) The function 'write' overrides a function in a base class but is not marked with a 'override' specifier.
[drivers\Linux\Print.h:68] -> [drivers\Linux\SerialPort.h:94]: (style) The function 'write' overrides a function in a base class but is not marked with a 'override' specifier.
[drivers\Linux\Stream.h:54] -> [drivers\Linux\SerialPort.h:103]: (style) The function 'peek' overrides a function in a base class but is not marked with a 'override' specifier.
[drivers\Linux\Stream.h:55] -> [drivers\Linux\SerialPort.h:107]: (style) The function 'flush' overrides a function in a base class but is not marked with a 'override' specifier.
[*\MyHwNRF5.cpp]: (information) Unmatched suppression: unreadVariable
(information) Cppcheck cannot find all the include files (use --check-config for details)
Correct the errors until cppcheck passes using 'git cppcheck --cached'.

michael@michael-PC MINGW64 ~/Documents/Arduino/libraries/MySensors/hal/transport/HC12 (hc12Radio)

And I do not know how to fix it.

fallberg:
Nice work! 
(assistance given though the forum)

MySensors.h
```diff
@@ -273,8 +284,8 @@ MY_DEFAULT_RX_LED_PIN in your sketch instead to enable LEDs
 #define MY_TRANSPORT_SANITY_CHECK		//!< enable regular transport sanity checks
 #endif
 
-// TRANSPORT INCLUDES
-#if defined(MY_RADIO_RF24) || defined(MY_RADIO_NRF5_ESB) || defined(MY_RADIO_RFM69) || defined(MY_RADIO_RFM95) || defined(MY_RS485)
+// TRANSPORT INCLUDES //Added MY_RADIO_HC12
```
tbowmo: unnecessary comment

MyConfig.h
```diff
@@ -249,6 +249,102 @@
...
+/**
+ * @defgroup HC12SettingGrpPub HC12
+ * @ingroup RadioSettingGrpPub
+ * @brief These options are specific to the HC12 serial radio transport.
+ * @{
+ */
+
+/**
+ * @def MY_HC12
+ * @brief Define this to use the HC12 wired transport for sensor network communication.
```
mfalkvidd: This brief is a bit confusing to me. Is the HC12 a radio or a wired transport?

MyConfig.h
```diff
@@ -249,6 +249,102 @@
...
+ * @brief The maximum message length used for HC12.
+ */
+#ifndef MY_HC12_MAX_MESSAGE_LENGTH
+#define MY_HC12_MAX_MESSAGE_LENGTH (40)
+#endif
+
+
+/**
+ * @def MY_HC12_SOH_COUNT
+ * @brief Use this in case of collisions on the bus. 3 might be a good setting.
```
mfalkvidd: If 3 is a good setting, should it be the default?

MyConfig.h
```diff
@@ -249,6 +249,102 @@
...
+ */
+#ifndef MY_HC12_SOH_COUNT
+#define MY_HC12_SOH_COUNT (1)
+#endif
+
+/**
+ * @def MY_HC12_SET_PIN
+ * @brief Define this to change the chip enable pin from the default.
```
mfalkvidd: "chip enable pin"->"set pin", to be consistent with the name (MY_HC12_SET_PIN)

MyConfig.h
```diff
@@ -249,6 +249,102 @@
...
+ */
+
+#ifndef MY_HC12_SET_PIN
+#if defined MY_RADIO_HC12
+//#warning SET is not defined, Remember to set before include of mysensor.h if needed
```
mfalkvidd: SET->MY_HC12_SET_PIN, otherwise people might try to define SET which won't do any good.
"mysensor.h" -> "MySensors.h"

MyConfig.h
```diff
@@ -249,6 +249,102 @@
...
+#endif
+#endif
+
+/**
+ * @def MY_HC12_TXPIN
```
mfalkvidd: MY_HC12_TXPIN -> MY_HC12_TX_PIN to use consistent naming with MY_HC12_SET_PIN
Perhaps it should be called MY_HC12_TXD_PIN (TX->TXD) because that's what it is called in the datasheet and on the schematics diagram at https://www.14core.com/wiring-the-hc11-hc12-434433mhz-transceiver/

MyConfig.h
```diff
@@ -249,6 +249,102 @@
...
+#if defined(__AVR_ATmega168__) ||defined(__AVR_ATmega168P__) ||defined(__AVR_ATmega328P__)
+#define MY_HC12_TXPIN (9)
+#else
+#define MY_HC12_TXPIN (14)
+#endif
+#endif
+
+
+/**
+ * @def MY_HC12_RXPIN
```
mfalkvidd: MY_HC12_RXPIN -> MY_HC12_RXD_PIN - see earlier comment

MyConfig.h
```diff
@@ -2298,6 +2395,12 @@
 #define MY_RF24_ENABLE_ENCRYPTION
 #define MY_RX_MESSAGE_BUFFER_FEATURE
 #define MY_RX_MESSAGE_BUFFER_SIZE
+// HC12
+#define MY_HC12
+#defien MY_HC12_MAX_MESSAGE_LENGTH
```
mfalkvidd: defien -> define

MySensors.h
tipih: Sorry, just forgot to remove it

tipih:
Hi all.

I would be very glad to fix all the issues listed above, but for now Im out of ideas on how to get this into the system.

Situation is like this.
When compiling for all the different Arduino platform it uses AltSoftwareSerial, that is already included in the build system, but but when compiling for the more highend platforms like the ESP32 ESP8266 and so on it was using the SoftwareSerial, currently included in the standard Arduino installation, but as it should be independent of external libraries I created a espsoftwareserial driver, and included the code for the esp software serial, BUT, that is using a lot of depensies to the 8266 platform, and currently the build system do not see all thowse. 
I think it is to big a job to include every single file from the ESP core, so therefor if anybody have idea on how this can be solve, I would gladly try it out.
btw.
It compiles fine on my local mashine, and I only have a standard Arduino installation no external libraries except for mySensor.

Br.
Michael

MyConfig.h
tipih: Thanks

MyConfig.h
tipih: I runs fine, but cannot be configured, meaning it will run with was it was manual configured with, so if power was set to P1 it will not be changed

MyConfig.h
tipih: Agree

mfalkvidd:
I agree that adding the esp8266 stuff into the MySensors library is troublesome. It would cause serious problems for users if the esp8266 core updated, making them incompatible with MySensors' copy. And people might use different versions of the esp8266 core and different versions of MySensors.

I don't know enough about the Jenkins environment works so I can't suggest any way forward unfortunately.

tipih:
Maybe it could be possible to to have the build system to include the esp core, as this is a project one it's own, if not then I don't think it will be possible to add low-level esp functionality.

tipih:
I still think being independent of libraries is fine, but core stuff is very tricky, and should be avoided, I think.

Br.
Michael

tipih:
Hi people.

What is your opinion, should I close the pull request, or keep it open ?

mfalkvidd:
Unless you are in extreme hurry to close it, I would say keep it open to give people a chance to respond.

tipih:
Absolutely, I just wanted to hear, I really think it would be a good idea to solve this issue, as it open up for other things as well.

tipih:
Adding support for HC12 radio is possible, but with the current build system, does not allow to use functions from the base ESP core. If the build system will support this later on, I will gladly do a new pull request, but for now I will close this request, but I will maintain a fork for the HC12 at my site.

Br.
Michael

PR #1179, opened 2018-08-14
Update RF24.cpp

Berkseo:
https://forum.mysensors.org/topic/9598/smartsleep-wake-up-message-always-get-nack/11


mfalkvidd:
Nice work. Would you mind adding a comment explaining why a longer time than specified in the datasheet is used? Otherwise, someone might change it back in the future thinking the time is unnecessarily long.

tekka007:
@Berkseo The NACKs after smartsleep wakeup is due to a missing delay() after powering-up the RF24 radio. This has been addressed in #1193 (see references therein).

Issue #1134, opened 2018-05-20
Software ack reponse too quick on nRF24


Yveaux:
**MySensors Version:** 2.2.0
**Hardware:** 
* STM32 BluePill @72MHz (Serial) Gateway + any type of node
* nRF24 radios

**What I'm doing:** Bench marking different gateways (currently a STM32 Serial & Ehternet MQTT)

**Problem Encountered:** Sending loads of messages from the node to the gateway. Each message send from the node requests an ack'd from the gateway.

* When I Enable  *#define MY_DEBUG* in the sketch everything works flawlessly and high message rates can be achieved without any loss of hardware (the automatic one from the nRF24 radio) or software acknowledge from the gateway.
* When I Remove  *#define MY_DEBUG* in the sketch I suddenly have loads of messages which do not appear to receive a hardware or software acknowledge from the gateway.

**Expected Cause:**
* It appears that the automatic hardware ack from the gateway radio gets too little time to complete due to the requested software acknowledge to be transmitted.  I.e. without the additional debug messages (=delay) present in the code a high clock speed microcontroller (like the STM32) is too fast in pushing a software ack to the radio before it can properly complete the hardware ack message.

According to the nRF24 datasheet (section 7.9) there must at least be 130us between the completion of the transmitted message and the hardware ack. response. On top of the 130us we also need some time to transmit the actual ack message. This total time must at < ARD (Auto Retransmit Delay), which is configures to be 1500us for MySensors.

So in **MyTransport.cpp** in *transportProcessMessage* just before the (void)transportSendRoute(_msgTmp) I added a delay of 1500uS and it all worked reliably again (successful hardware and software ack's and significant reduction of auto re-transmits of 16 times). This delay can actually be smaller than the 1500us as there is of course some time consumed by all the other tasks (still appeared to work properly with only 150us).

Swiftnesses:
This explains some of my issues, hope it can be implemented soon!

cimba007:
Just for reference .. I had a similar timing issue with a fast gateway and slow nodes

https://github.com/mysensors/MySensors/issues/578

mfalkvidd:
Nice work Yveaux.

This seems to be the relevant part of the datasheet:
![image](https://user-images.githubusercontent.com/893502/57586858-2ef68800-74fc-11e9-9998-6285b7a5a16c.png)

[...]

PR #1180, opened 2018-08-14
Update RF24registers.h

Berkseo:
https://forum.mysensors.org/topic/9598/smartsleep-wake-up-message-always-get-nack/11

### DIFF
--- a/drivers/RF24/RF24registers.h
+++ b/drivers/RF24/RF24registers.h
@@ -41,7 +41,7 @@
 #define RF24_CRC_16				(3)
 
 // ARD, auto retry delay
-#define RF24_SET_ARD			(5)	//=1500us
+#define RF24_SET_ARD			(10)	//=1500us
 
 // ARD, auto retry count
 #define RF24_SET_ARC			(15)


mfalkvidd:
Could the problem be realted to https://forum.mysensors.org/topic/9642/nrf24-transmission-of-data-works-fine-but-constant-nack-s-produced/ so the fix suggested there is a better solution?

Berkseo:
Increasing the number of attempts is aimed at correcting this line - 2912 !TSF:MSG:SEND,100-100-0-0,s=255,c=3,t=26,pt=1,l=1,sg=0,ft=0,st=NACK:2.

This does not apply to NOACK in the first message after waking up

As I understand in this the frost is offered to exclude delay in 10ms(RF24.cpp). I think it's the wrong decision. The specification says it explicitly - at least 10 MS (>=10). For the original radio modules, a 10ms delay is probably sufficient to output the signal to the output. For most radio modules, this is clearly not enough.

drivers/RF24/RF24registers.h
```diff
@@ -41,7 +41,7 @@
 #define RF24_CRC_16				(3)
 
 // ARD, auto retry delay
-#define RF24_SET_ARD			(5)	//=1500us
+#define RF24_SET_ARD			(10)	//=1500us
```
mfalkvidd: Is the comment (1500us) still valid after this change?

Berkseo:
no, I'm sorry, I just did not remove the comment, of course there will be more time.

Yveaux:
This solution is likely to conflict with #1134.

tekka007:
see #1193

PR #1197, opened 2018-08-30
Inclusion mode: change pinmode to INPUT_PULLUP

### LINKED ISSUE #1196: Inclusion mode button does not work with ESP32
After testing the inclusion mode with an ESP32 gateway I noticed the inclusion button does not work. 

The pin is in floating state instead of pulled up.
Init code:
https://github.com/mysensors/MySensors/blob/6d24ab1bb52ed4402140def8f5ba83fa51b7c0d2/core/MyInclusionMode.cpp#L33

According to [arduino references](https://www.arduino.cc/en/Tutorial/DigitalPins) we should instead use INPUT_PULLUP as of Arduino 1.0.1

I will submit a PR for this.

freynder:
- Fixes #1196

According to arduino references we should instead use INPUT_PULLUP as of Arduino 1.0.1
https://www.arduino.cc/en/Tutorial/DigitalPins

### DIFF
--- a/core/MyInclusionMode.cpp
+++ b/core/MyInclusionMode.cpp
@@ -30,8 +30,7 @@ inline void inclusionInit()
 	_inclusionMode = false;
 #if defined(MY_INCLUSION_BUTTON_FEATURE)
 	// Setup digital in that triggers inclusion mode
-	hwPinMode(MY_INCLUSION_MODE_BUTTON_PIN, INPUT);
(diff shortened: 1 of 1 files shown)


tekka007:
@freynder The cited Arduino reference is specific to atmega, does this also apply to other architectures?

freynder:
@tekka007 Another source is the [pinMode](https://www.arduino.cc/en/Reference.PinMode) reference which does not mention any particular architecture and explicitely says INPUT disables internal pullups. I can confirm it works for ESP32 whereas INPUT did not work.

PR #1198, opened 2018-08-30
Allow flash strings in present() to save RAM

### LINKED ISSUE #1194: Allow F() macro on child names during presentation
Currently this is a normal present call:

_present(SMS_CHILD_ID, S_INFO, "received sms");_

But it would be great it this was also possible, as it would save memory:

_present(SMS_CHILD_ID, S_INFO, F("received sms"));_

trying it gives an error: 
`cannot convert 'const __FlashStringHelper*' to 'const char*' for argument '3' to 'bool present(uint8_t, uint8_t, const char*, bool)'`

The F macro does already work here (since MySensors 2.2 I believe):
_sendSketchInfo(F("SMS & password relays"), F("1.62"));_

There is also a forum thread about this:
https://forum.mysensors.org/topic/9653/using-the-f-macro-in-the-presentation-function

tekka007:
- Fixes #1194

### DIFF
--- a/core/MySensorsCore.cpp
+++ b/core/MySensorsCore.cpp
@@ -355,8 +355,19 @@ bool present(const uint8_t childSensorId, const uint8_t sensorType, const char *
              const bool ack)
 {
 	return _sendRoute(build(_msgTmp, GATEWAY_ADDRESS, childSensorId, C_PRESENTATION, sensorType,
-	                        ack).set(childSensorId==NODE_SENSOR_ID?MYSENSORS_LIBRARY_VERSION:description));
+	                        ack).set(childSensorId == NODE_SENSOR_ID ? MYSENSORS_LIBRARY_VERSION : description));
 }
+
+#if !defined(__linux__)
+bool present(const uint8_t childSensorId, const uint8_t sensorType,
+             const __FlashStringHelper *description,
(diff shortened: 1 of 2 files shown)


core/MySensorsCore.cpp
```diff
@@ -355,9 +355,20 @@ bool present(const uint8_t childSensorId, const uint8_t sensorType, const char *
              const bool ack)
 {
 	return _sendRoute(build(_msgTmp, GATEWAY_ADDRESS, childSensorId, C_PRESENTATION, sensorType,
-	                        ack).set(childSensorId==NODE_SENSOR_ID?MYSENSORS_LIBRARY_VERSION:description));
+	                        ack).set(childSensorId == NODE_SENSOR_ID ? MYSENSORS_LIBRARY_VERSION : description));
 }
 
+#if !defined(__linux__)
```
mfalkvidd: Is there no clean way to avoid the code duplication? I worry that a change in one of these nearly identical functions might be done without making the corresponding change to the other function.

tekka007: The magic happens in MyMessage::set() - I do not see how this can be improved.

mfalkvidd: ok

Issue #1057, opened 2018-02-07
Splash screen not redirected to MY_DEBUGDEVICE


tekka007:
See forum:
https://forum.mysensors.org/topic/8989/splash_screen-with-my_debugdevice

nelsonov:
I have a fix for this.  It's only a couple lines in MySplashScreen.cpp.  However, in working on this I uncovered something related.  displaySplashScreen is called before the sketch's setup().  If someone is using a soft serial port, that port will not be started until after the splash screen is displayed.  This makes the splash screen disappear into a black hole.

Should I start a PR for this issue, #1057, as originally stated?

Should I create a separate issue for the problem of the order of displaySplashScreen and setup()?

nelsonov:
As noted in #1207 , if using a software serial device as MY_DEBUGDEVICE, then it needs to initialized in preHwInit() for the splash screen to work.  Later functions like before() and setup() happen after the splash screen has already been displayed.

```
void preHwInit()
{
  SoftSerial.begin(9600);
}
```

PR #1210, opened 2018-10-09
Allow splash screen to redirect to MY_DEBUGDEVICE if defined (#1057)

nelsonov:
This addresses issue #1057


core/MySplashScreen.cpp
```diff
@@ -19,6 +19,10 @@
 
 #include "MySensorsCore.h"
 
+#ifndef MY_DEBUGDEVICE
+#define MY_DEBUGDEVICE MY_SERIALDEVICE
```
tekka007: Not needed, definition set in arch headers (MyHw****.h)

PR #1241, opened 2018-12-28
Allow definition of MY_OTA_RETRY and MY_OTA_RETRY_DELAY

pragtich:
Hi there,

Playing with OTA firmware update, and struggling a bit because of the small number of retries and short retry delay. Essentially, a radio disturbance during 2.5s can lead to a failure of the OTA. This caused most uploads to fail in my situation.

Would you consider allowing the user to change MY_OTA_RETRY and/OR MY_OTA_RETRY_DELAY in order to influence their OTA process?


pragtich:
FYI, I had an issue with failed checks due to a coding style problem (I had an indented #define).
The butler didn't manage to get an email to me: that made it a bit difficult to guess what was going on :-)

Should be good now...

tekka007:
@pragtich thanks for your contribution!

PR #1248, opened 2019-01-04
Add MQTT Wifi connection status check

### LINKED ISSUE #788: Restart of Gateway on lost of WiFi on ESP8266
I have noted that if the Access Point that the MQTT-ESP8266 gateway  is connect to is turned off and re powered on. IE: a power failure

That the gateway will not re connect to the AP without a re-boot...

MySensor 2.1.1
IDE 1.8.1

Yveaux:
Fix #788, causing crash on ESP8266 when Wifi connection to AP fails with MQTT connected.

### DIFF
--- a/core/MyGatewayTransportMQTTClient.cpp
+++ b/core/MyGatewayTransportMQTTClient.cpp
@@ -273,6 +273,11 @@ bool gatewayTransportAvailable(void)
 	if (_MQTT_connecting) {
 		return false;
 	}
+#if defined(MY_GATEWAY_ESP8266) || defined(MY_GATEWAY_ESP32)
+	if (WiFi.status() != WL_CONNECTED) {
+		return false;
+	}
+#endif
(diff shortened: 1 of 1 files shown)


core/MyGatewayTransportMQTTClient.cpp
```diff
@@ -273,6 +273,11 @@ bool gatewayTransportAvailable(void)
 	if (_MQTT_connecting) {
 		return false;
 	}
+#if defined(MY_GATEWAY_ESP8266) || defined(MY_GATEWAY_ESP32)
+	if (WiFi.status() != WL_CONNECTED) {
+		return false;
```
tekka007: ESP32 won't reconnect unless you call gatewayTransportInit();

Yveaux: Ok, reworked based on your comments

PR #1253, opened 2019-01-12
Fixing NRF52 Transport "EEPROM_RF_ENCRYPTION_KEY_ADDRESS"

desander:
Changing EEPROM_RF_ENCRYPTION_KEY_ADDRESS to EEPROM_RF_ENCRYPTION_AES_KEY_ADDRESS


tekka007:
@desander good catch re encryption key address! What's the purpose of the changes in MyHwNRF5.cpp?

desander:
@tekka007 
I only changed the MyHwNRF5 because I had the impression that it made the readings more even. This is actually only for viewing in a graph. I didn't invest much time to find the optimal values. It could be that a 2X oversampling would also be enough. So if you are thinking about taking this change as well, you would have to find the optimal values first. Pro is that the measured values are more even. Contra is the extra time that a measured value needs and therefore the battery consumption increases minimally. Since in my node however only 1 battery reading per hour is sent it is in my case a clear decision for the uniformity.

desander:
Sorry i'am new on Github so...
It was not my intend to pull the Changes from MyHwNRF5.cpp. So how this is happend ...?

desander:
Okay so i have to create a branch on the fork so i can seperate changes. Those who going to merge to this repo and those who are primary for my personal preferences. I close this Pull Request and will open a new one when its done.

PR #1259, opened 2019-01-15
Fix not sending the messages to MQTT gateway with RF24

schalkje:
Based on a sugestion in the feedback; solves my problem that messages are not being sent to MQTT. The state changed were visible in the serial monitor. Strange

patrikr76 commented 6 months ago
So i was testing out the motionsensor sketch and it kept giving me "!TSF:MSG:SEND" after the initial 0 had been sent.

Finding no answers on here i tested loads of things until i tried a delay after the sleep command.
Running fine now with a 5msec delay.

Just thought i would share.


mfalkvidd:
Do you know if the problem was resolved by 
https://github.com/mysensors/MySensors/pull/1193 ?
If it was resolved - no need for this change
If it was not resolved - the problem will lost likely affect all examples, not only the motion sensor example.

schalkje:
Took me some time to get this verified. Fixed in 2.3.1

mfalkvidd:
Great, thanks for confirming!

PR #1268, opened 2019-02-17
Implementation of message queue for RFM69 new driver

koreshx:
The changes are affected the mainline only in case of new rem69 driver is used and MY_RX_MESSAGE_BUFFER_FEATURE is defined.


hal/transport/RFM69/driver/new/RFM69_new.cpp
```diff
@@ -384,7 +443,16 @@ LOCAL bool RFM69_sendFrame(rfm69_packet_t *packet, const bool increaseSequenceCo
 	while (!RFM69_channelFree() &&
 	        ((hwMillis() - CSMA_START_MS) < MY_RFM69_CSMA_TIMEOUT_MS)) {
 		doYield();
+		
+		//A little delay before the next attempt
+		delay(1);
```
mfalkvidd: Is 1ms the right amount of time to delay? Could we document why 1ms was chosen?

hal/transport/RFM69/driver/new/RFM69_new.cpp
```diff
@@ -565,6 +644,69 @@ LOCAL bool RFM69_standBy(void)
 }
 
 
+LOCAL void RFM69_sendACK_fast(const uint8_t recipient, const rfm69_sequenceNumber_t sequenceNumber,
```
mfalkvidd: When should RFM69_sendACK_fast be used and when should RFM69_sendACK be used? Is there a reason both need to exist?

hal/transport/RFM69/driver/new/RFM69_new.h
```diff
@@ -150,7 +150,7 @@
 
 // CSMA settings
 #if !defined(MY_RFM69_CSMA_LIMIT_DBM)
-#define MY_RFM69_CSMA_LIMIT_DBM             (-95)			//!< upper RX signal sensitivity threshold in dBm for carrier sense access
+#define MY_RFM69_CSMA_LIMIT_DBM             (-85)			//!< upper RX signal sensitivity threshold in dBm for carrier sense access
```
mfalkvidd: Could we document why -85 is better than -95?

MyConfig.h
```diff
@@ -354,7 +354,7 @@
  * This feature is currently not supported for anything but RF24.
  * Require @ref MY_RF24_IRQ_PIN to be set.
  */
-//#define MY_RX_MESSAGE_BUFFER_FEATURE
+#define MY_RX_MESSAGE_BUFFER_FEATURE
```
mfalkvidd: Won't defining this by default break a lot of existing installations (that don't have MY_RF24_IRQ_PIN defined, or don't have the irq pin connected)?

MyConfig.h
```diff
@@ -364,7 +364,7 @@
  */
 #ifdef MY_RX_MESSAGE_BUFFER_FEATURE
 #ifndef MY_RX_MESSAGE_BUFFER_SIZE
-#define MY_RX_MESSAGE_BUFFER_SIZE (20)
+#define MY_RX_MESSAGE_BUFFER_SIZE (10)
```
mfalkvidd: Could we document why 10 is a better value than 20?

hal/transport/MyTransportHAL.h
```diff
@@ -29,9 +29,9 @@
 #if defined(MY_RADIO_NRF5_ESB)
 #error Receive message buffering not supported for NRF5 radio! Please define MY_NRF5_RX_BUFFER_SIZE
 #endif
-#if defined(MY_RADIO_RFM69)
-#error Receive message buffering not supported for RFM69!
-#endif
+//#if defined(MY_RADIO_RFM69)
```
mfalkvidd: If this is not needed, just remove it. We don't want clutter from unused code.

hal/transport/RFM69/driver/new/RFM69_new.cpp
koreshx: I tried to not touch current rfm69 new driver workflow. So if MY_RX_MESSAGE_BUFFER_FEATURE is not defined old logic works. If MY_RX_MESSAGE_BUFFER_FEATURE is defined workflow is differ. In this case incoming messages are processed by interrupt. Ack messages are sent inside interrupt too (RFM69_sendACK_fast) so there is not possible to freeze the thread or use time functions.

hal/transport/RFM69/driver/new/RFM69_new.cpp
koreshx: I saw some CSMA realisations and all of them contains such delay. There is no reason check the air immediately after it is busy. Moreover it is nice to implement randomized delay... 1ms works fine in my tests but I do not insist on exacly this value.

hal/transport/RFM69/driver/new/RFM69_new.h
koreshx: It will improve traffic in case of very noisy power supply (usb ftdi adapter with ultra low power ldo for example :)).  Of course it increases the probability of collisions in case of very long range. We can omit this but put the comment.

MyConfig.h
koreshx: Sure.

hal/transport/MyTransportHAL.h
koreshx: Ok

hal/transport/RFM69/driver/new/RFM69_new.cpp
mfalkvidd: 👍 Thanks for clarifying.

MyConfig.h
koreshx: MAX_MESSAGE_LENGTH is 32. So MAX_MESSAGE_LENGTH *20=640 ~1/3 of the meemory. I don't think so large queue is needed for nodes except the gate or repeater. May be reasonable to change this definition according the used platform.

hal/transport/RFM69/driver/new/RFM69_new.cpp
mfalkvidd: Sounds reasonable. Let's stick with the current value unless anyone else has an opinion.

hal/transport/RFM69/driver/new/RFM69_new.h
mfalkvidd: I just remembered your comment in 
https://forum.mysensors.org/post/91435
I suggest we go with -85 as default value, since your tests show that -95 is too optimistic.

MyConfig.h
mfalkvidd: I agree that 10 should be sufficient for most nodes. Changing it based on platform would be nice, but maybe out of scope for this pull request (unless you feel like adding it).

MyConfig.h
mfalkvidd: Maybe define it by default for RFM but not for NRF? RFM will always have irq connected anyway, right? So only drawback is that the buffer uses more ram.

koreshx: Maybe for new rfm69 driver. New rfm69 driver is not often used in real setups I think.

Yveaux: Agree with @mfalkvidd. This breaks NRF24 setups without IRQ pin connected and might influence timing for NRF24 setups.
Don't make it default in MyConfig.h.

Yveaux: Also, comments say this only works for NRF24, which isn't true anymore after merging your PR ;-)

drivers/RFM69/new/RFM69_new.cpp
```diff
@@ -239,9 +248,58 @@ LOCAL void RFM69_clearFIFO(void)
...
 	// set flag
-	RFM69_irq = true;
+	//RFM69_irq = true;
+
+
+	if (RFM69_receiveCallback) {
+		// Start checking if RX-FIFO is not empty, as we might end up here from an interrupt
+		// for a message we've already read.
+		//while (RFM69_available()) {
+			RFM69_receiveCallback();		// Must call RF24_readMessage(), which will clear RX_DR IRQ !
```
Yveaux: Update the comments, remove commented code

drivers/RFM69/new/RFM69_new.cpp
```diff
@@ -297,6 +355,8 @@ LOCAL void RFM69_interruptHandling(void)
 						RFM69.currentPacket.payloadLen = readingLength;
 						RFM69.ackReceived = RFM69_getACKReceived(RFM69.currentPacket.header.controlFlags);
 						RFM69.dataReceived = !RFM69.ackReceived;
+
+						RFM69_DEBUG(PSTR("RFM69:MH:l=%" PRIi8 " pl=%" PRIi8 "\n"), readingLength, (RFM69.ackReceived ? 1 : 0));
```
Yveaux: You're not support to serial print from interrupts

drivers/RFM69/new/RFM69_new.cpp
```diff
@@ -378,7 +440,18 @@ LOCAL bool RFM69_sendFrame(rfm69_packet_t *packet, const bool increaseSequenceCo
 	while (!RFM69_channelFree() &&
 	        ((hwMillis() - CSMA_START_MS) < MY_RFM69_CSMA_TIMEOUT_MS)) {
 		doYield();
+
+		//KORESH,  A LITTLE DELAY BEFORE NEXT ATTEMPT
```
Yveaux: Don't prefix your comments with //KORESH please

hal/transport/RFM69/driver/new/RFM69_new.cpp
```diff
@@ -244,9 +251,59 @@ LOCAL void RFM69_clearFIFO(void)
...
+	const uint8_t sender = RFM69.currentPacket.header.sender;
+	const rfm69_sequenceNumber_t sequenceNumber = RFM69.currentPacket.header.sequenceNumber;
+	const uint8_t controlFlags = RFM69.currentPacket.header.controlFlags;
+	const rfm69_RSSI_t RSSI = RFM69.currentPacket.RSSI;
+
+	if (buf != NULL) {
+		(void)memcpy((void *)buf, (void *)&RFM69.currentPacket.payload, payloadLen);
+	}
+
+	RFM69_DEBUG(PSTR("RFM69:RM:s=%" PRIi8 " pl=%" PRIi8 "\n"),sender, payloadLen);
```
Yveaux: Called from interrupt context; no serial printing allowed.

Yveaux:
In general, how did you test your changes?

drivers/RFM69/new/RFM69_new.cpp
koreshx: These comments should be deleted in the latest version as I remember.

drivers/RFM69/new/RFM69_new.cpp
koreshx: ok

hal/transport/RFM69/driver/new/RFM69_new.cpp
koreshx: ok

koreshx:
> In general, how did you test your changes?

 I tested this version long tyme ago with case described here: https://forum.mysensors.org/topic/9484/mysensors-2-3-and-rfm69-new-driver-problems-and-solutions And queue system solved all collisions. As I see that problem is not actual in current mysensors version. Also Me and Yury try some setups in abnormal conditions last month. All nodes floods enough but the system works nice. May be this implementation need more tests and improovments. But I am sure it is much more reliable then rfm69_new_driver basic implementation (which is not default for mysensrs yet)

mfalkvidd:
Today I tested this on my RMF69 node.

Without buffer feature:
Sketch uses 24648 bytes (80%) of program storage space. Maximum is 30720 bytes.
Global variables use 1132 bytes (55%) of dynamic memory, leaving 916 bytes for local variables. Maximum is 2048 bytes.

With buffer feature:
Sketch uses 25312 bytes (82%) of program storage space. Maximum is 30720 bytes.
Global variables use 1471 bytes (71%) of dynamic memory, leaving 577 bytes for local variables. Maximum is 2048 bytes.

So the cost is 664 bytes flash and 339 bytes RAM. I guess that's reasonable.

The node seems to work like before. I don't have much traffic so I didn't expect it to work differently, but the communication seems to be at least as good as before.

mfalkvidd:
When compiling (for both gateway and node) I get the following warnings. The wartnings are not present when compiling with upstream/development.
```
In file included from R:\...\MySensors/MySensors.h:376:0,

                 from C:\...\arduino_modified_sketch_784083\Arduino-MySensors-RFM69StorageRoom.ino:18:

R:\...\MySensors/hal/transport/RFM69/driver/new/RFM69_new.cpp:380:12: warning: 'RFM69_handler' defined but not used [-Wunused-function]

 LOCAL void RFM69_handler(void)

            ^

R:\...\MySensors/hal/transport/RFM69/driver/new/RFM69_new.cpp:405:15: warning: 'RFM69_receive' defined but not used [-Wunused-function]

 LOCAL uint8_t RFM69_receive(uint8_t *buf, const uint8_t maxBufSize)
```

mfalkvidd:
When using this branch on my gateway (regardless if the buffer feature is enabled or not), the gateway hangs after a while. The serial debug looks like this
image
It would be a coincidence, but so far it is reproducible (tested as failing 5 times in a row on this branch and working 5 times in a row using upstream/development).

image
image
image
image

mfalkvidd:
I have analyzed the logs and my impression is that the corrupted serial prints happens when the gateway is trying to send an ACK, waiting for CSMA, and the node sends a new message.

hal/transport/RFM69/driver/new/RFM69_new.cpp
```diff
@@ -384,7 +443,16 @@ LOCAL bool RFM69_sendFrame(rfm69_packet_t *packet, const bool increaseSequenceCo
 	while (!RFM69_channelFree() &&
 	        ((hwMillis() - CSMA_START_MS) < MY_RFM69_CSMA_TIMEOUT_MS)) {
 		doYield();
+		
+		//A little delay before the next attempt
+		delay(1);
+	}
+
```
mfalkvidd: Earlier behavior was to send anyway if the CSMA time limit was reached. If the new behavior is used, messages will not be sent at all in environments like mine (using the default settings). See https://forum.mysensors.org/topic/10224/csma-lbt-for-rfm69-new-driver/
Not sure what is best here. I really want the library to work by default for most users. But sending if there is a lot of noise isn't nice, so in that aspect the new behavior is better.

koreshx:
> I have analyzed the logs and my impression is that the corrupted serial prints happens when the gateway is trying to send an ACK, waiting for CSMA, and the node sends a new message.

Thanks for the great job! Yes, it is necessary to remove all serial prints inside the interrupt handler. Unfortunately I have no time right now because I am prepearing one of hardware project to crowdfund. But I will end this work soon, hope.

hal/transport/RFM69/driver/new/RFM69_new.cpp
mfalkvidd: If we opt for blocking the message, there needs to be a very clear log message (visible on normal debug level or even higher if possiblej. Otherwise nodes will just silently drop all outgoing messages and people will be very confused.

tekka007:
stale PR

PR #1270, opened 2019-03-18
smartSleep revokation feature

tmandel123:
Revoke going to sleep if there is a new C_SET Command after sending I_PRE_SLEEP_NOTIFICATION


tmandel123:
have to resolve "Tall gate" error

PR #1271, opened 2019-03-18
smartSleep revocation feature

tmandel123:
Revoke going to sleep if there is a new C_SET Command after sending I_PRE_SLEEP_NOTIFICATION


tekka007:
@tmandel123 Great idea, how about taking this into consideration: Adding either a dedicated sleep revoke message type, i.e. C_SET/I_REVOKE_SLEEP or (even better) implementing a wait() function detecting specific C_ commands?

tekka007 (changes requested):

core/MySensorsCore.cpp
```diff
@@ -638,7 +638,18 @@ int8_t _sleep(const uint32_t sleepingMS, const bool smartSleep, const uint8_t in
 		// notify controller about going to sleep, payload indicates smartsleep waiting time in MS
 		(void)_sendRoute(build(_msgTmp, GATEWAY_ADDRESS, NODE_SENSOR_ID, C_INTERNAL,
 		                       I_PRE_SLEEP_NOTIFICATION).set((uint32_t)MY_SMART_SLEEP_WAIT_DURATION_MS));
-		wait(MY_SMART_SLEEP_WAIT_DURATION_MS);		// listen for incoming messages
+		if (wait(MY_SMART_SLEEP_WAIT_DURATION_MS, C_SET,
+		         I_VERSION)) { //cancel sleeping request if there is a new C_SET Command
```
tekka007: This is not accurate as sleep is only revoked if C_SET/I_VERSION is received. The implementation of a generic `wait(ms, cmd)` function is required here.

tekka007 (changes requested):

core/MySensorsCore.cpp
tmandel123: What usually is the msgType if a node receives a new C_SET? What i saw was I_VERSION. Can C_SET have msgTypes other than I_VERSION?

tekka007: E.g. internal messages (see here https://github.com/mysensors/MySensors/blob/development/core/MyMessage.h#L166-L201). Actually, the more I think about the sleep revoke feature, a new message type `I_SLEEP_REVOKE` to specifically prevent sleeping makes the most sense.

tmandel123: This feature was discussed a little bit here:
[https://forum.mysensors.org/topic/10221/disable-cancel-smartsleep-while-sleep-countdown-is-running](url)

At the moment (at least the controller FHEM) sends queued messages to sleeping nodes when it receives I_PRE_SLEEP_NOTIFICATION. I my case the controller just sends "turn LED on" message.  If we create a new message type like I_SLEEP_REVOKE then the controller has to know what the send in advance. I want, that my node stays on (no sleeping) if the LED is on but when the LED is switched off the node goes to sleep for 30 seconds. The time of 30 seconds i short enough to "look for new commands" after issuing I_PRE_SLEEP_NOTIFICATION.

In my opinion the generic `wait(ms, cmd)` would be more compatible with existing controllers. 
Maybe you have other great plan for the future of mysensors.

tekka007:
stale PR

Issue #1242, opened 2018-12-30
License


ramack:
While browsing through the code I stumbled over the ["GPL v3 or later" statement of the SHA256 implementation for AVR](https://github.com/mysensors/MySensors/blob/03767e34ddf2882952382d672e4bc89711c241aa/hal/crypto/AVR/MyCryptoAVR.cpp#L41)
while the [mysensors code seems to be GPLv2-only](https://github.com/mysensors/MySensors/blob/03767e34ddf2882952382d672e4bc89711c241aa/hal/crypto/AVR/MyCryptoAVR.cpp#L17)

according my understanding GPLv2 and GPLv3+ are incompatible and therefore cannot be linked together.

henrikekblad:
Good observation. If you enable sha256 on AVR there will be a license clash. We'll have to discuss how to handle it in the core team (a bit slow over the holidays though).

ramack:
no hurry, IANAL and I will not charge you :smile:
Why not changing all to GPLv3+?

While you are at license topics, maybe you can answer my question about the CLA / dual licensing [that I raised in the forum](https://forum.mysensors.org/post/96131)

henrikekblad:
Haha.. we're not lawyers either ;) The CLA was put into place for situations just like this. E.g. we wanting move the MySensors library to a GPLv3 license without having to ask all previous contributors.
It also opens up a possibility for a company to embed the MySensors core into some proprietary hardware thingie by using a "commercial" license (hasn't happened yet).

Anyhow. I can conclude that licensing isn't super easy.  Wish I had a little OSS savvy lawyer friend somewhere.

ramack:
I'd like to meet that guy too, please 😃
I understand that you want to allow a commercial use, but then maybe it is better to directly select a license without copyleft, or explicitly dual license it from the beginning. Currently for me it has a slight smell, as I as a potential contributor would need to give rights to you without knowing you and without a clear view of what you will or will not do with this right.
The integration of external code is always problematic - with the current CLA and with dual licensing...

ramack:
any news about licensing?

ramack:
any news? It is not a good state to keep a free software project in a known license clash situation for nearly 4 years!

I strongly recommend to fix that!

PR #1278, opened 2019-04-19
Add PJON transport layer

### LINKED ISSUE #1171: Feature request: implement PJON protocol for wired communication
Wired communication is more secure and private by default. This protocol looks great.

https://forum.mysensors.org/topic/3977/transport-based-on-single-wire-pjon-protocol/17

tekka007:
- Based on #1230, credit to @4ib3r
- PJON library updated to 12.1
- Fixes #1171


henrikekblad:
@gioblu,  @4ib3r, @gryzli133, would have be great if you could help out to do some tests on this PR before it goes in...

gioblu:
Ciao @tekka007, @henrikekblad thank you very much for pushing this through, although studying in detail your license [here](https://www.mysensors.org/about/mysensors#copyright) I think including the PJON source within MySensors would cause a license compatibily issue with the "Open Source Licensing GPL V2", not sure about the "Commercial Licensing" which clause is not public.

Following the stands of the Apache Foundation:
>Despite our best efforts, the FSF has never considered the Apache License to be compatible with GPL version 2, citing the patent termination and indemnification provisions as restrictions not present in the older GPL license. The Apache Software Foundation believes that you should always try to obey the constraints expressed by the copyright holder when redistributing their work.

Also following this discussion [here](https://opensource.stackexchange.com/questions/1357/can-i-link-a-apache-2-0-library-into-software-under-gplv2) seems it is considered incompatible even the mere linking of an Apache 2.0 library within a GPL v2.0 project.

From the Apache clause:
>You may add Your own copyright statement to Your modifications and may provide additional or different license terms and conditions for use, reproduction, or distribution of Your modifications, or for any such Derivative Works as a whole, **provided Your use, reproduction, and distribution of the Work otherwise complies with the conditions stated in this License**.

For you to switch to LGPLv3 may be favorable because APL licensed projects would be compatible and could be linked in MySensors, or externally use the PJON library in MySensors, or even include the source of PJON with its own APL license if I am not mistaken, see [here](http://www.apache.org/licenses/GPL-compatibility.html):

>Apache 2 software can therefore be included in GPLv3 projects, because the GPLv3 license accepts our software into GPLv3 works. However, GPLv3 software cannot be included in Apache projects. The licenses are incompatible in one direction only, and it is a result of ASF's licensing philosophy and the GPLv3 authors' interpretation of copyright law.
 
quick-guide-gplv3-compatibility

Consider also that external linking of third party libraries (and not their inclusion in a subdirectory of MySensors as it is now, that makes you a distributor) would let you have more freedom in using external tools with less chances to stumble in a license compatibility issue like this.

henrikekblad:
Yeah, the licensing is a bit of a mess. The best solution I've found so far (when doing some googling previously) seems to be to make MySensors available under 3 licenses GPLv2, GPLv3 and LGPLv3. A popular library, QT, made this switch some [a couple of years ago](https://blog.qt.io/blog/2016/01/13/new-agreement-with-the-kde-free-qt-foundation/). 

We already have an ongoing issue on the license problems here: https://github.com/mysensors/MySensors/issues/1242

Actpohomoc:
I asked for an expert on licenses, and he suggested looking at this article: http://www.apache.org/licenses/GPL-compatibility.html

gioblu:
Ciao @henrikekblad and @Actpohomoc thank you for your feedback.
I hope this will get resolved not only to let this pull-request pass through, but also to simplify the integration of other non-GPL tools in the future.

gryzli133:
What is the next step? Will PJON will be available in next Release of MySensors? :-)

tekka007:
@gryzli133 The license issue is currently under discussion and we will proceed with the PR ASAP. In the meantime, we need ppl testing this PR and providing feedback.

gryzli133:
How can I download the version that includes PJON? Sorry for noob questions, but I completely skipped git etc while do PLC programming

gioblu:
Ciao @gryzli133 I think you should download [98d3d33](https://github.com/mysensors/MySensors/tree/98d3d33ae98105fb0cc695d5ee8396d712ab8707) and give it a try :)

gioblu:
Ciao @tekka007 thank you for pushing this forward, @Swiftnesses @gryzli133 would you give a try to [b24711f](https://github.com/mysensors/MySensors/tree/b24711fdbe2e482867574abc44dc258bd3884e60)?

gryzli133:
Hi @gioblu, I will give it a try at the end of this week ;-)

gryzli133:
I have tested the PJON transport layer and it looks very promising. I have been using 98d3d33. The only one problem I found is when we disconnect the NODEs from each other, then the sending node hangs after trying to send the data. Is there any possibile way to say, that the controller should first of all do the logic and just send the data in the background? In my case the local logic should have the priority and sending data is a "nice to have feature" ;-)

Here video on YouTube: 
https://www.youtube.com/watch?v=P85FlQKpM8s

Here the sketches for Gateway and Node:
https://github.com/gryzli133/MySensors_PJON_test

gioblu:
CIao @gryzli133 thank you very much for your tests and your video review. I do not see how the hang issue can be caused by the PJON library itself. Testing bare PJON this does not happen. I suspect there is still something wrong in the MySensors's middleware.

gryzli133:
Is there any acknowledgement for the message? Or maybe I need to change connection timeout?

tekka007:
@gryzli133 Thanks for testing 👍, the MySensors PJON transport layer can print additional debug info by enabling `MY_DEBUG_VERBOSE_PJON`. Please post the full debug log for further troubleshooting.

gioblu:
Hei @gryzli133 quickly reading the source of MySensors I do see a mistake here:
https://github.com/mysensors/MySensors/commit/98d3d33ae98105fb0cc695d5ee8396d712ab8707#diff-f93d9576c0213eaef64d26e97cc5139aR48 at line 45 and 48.

`send_packet_blocking` is a blocking function and already does comply with the back-off so it does already retry by itself sending in this case 20 times:
https://github.com/gioblu/PJON/blob/master/src/strategies/SoftwareBitBang/Timing.h#L287

so that `while` at line 45 should not be there and because of that it will likely hung for 4 seconds (the maximum duration of `send_packet_blocking` if all transmissions retry fail) multiplied by `MY_PJON_MAX_RETRIES`

I do see 2 ways to solve the issue:
- Remove the `while` just use `send_packet_blocking` and monitor its result
- change `send_packet_blocking` in `send_packet` who tries only once to send a packet

@gryzli133 it may have sense to give a try to the latest code, I just have looked into https://github.com/mysensors/MySensors/commit/98d3d33ae98105fb0cc695d5ee8396d712ab8707

gryzli133:
I have tested it again with 98d3d33. Still the same problem, but this time I can show you debug (first few clicks while connected, then one try while disconnected - waiting until finished/timeout, then connecting the wire again while "hanging"):
https://github.com/gryzli133/MySensors_PJON_test/blob/master/LogGateway
https://github.com/gryzli133/MySensors_PJON_test/blob/master/LogNode1

If you find way to get it work, then just let me know, I will try to test it asap. Thank you guys for your support!

gryzli133:
@gioblu Is there a possibility, that the function "transportSend" will be working in Background? Or let say, instead of using while, just try once, then just wait for next loop (while user software will still work), then try again and so on.

gioblu:
Ciao @tekka007 PJON 12.0 has been released it may have sense to base this pull upon the newly released code.

Ciao @gryzli133 the function `send_packet` can be used to send the packet once, or `send_packet_blocking` to try to send the packet multiple times with back-off. If you want to schedule a packet sending and let PJON do the job you can use `send` and call `update` in the loop although this consumes a little more memory.

gryzli133:
Thank you for your support guys ! :-)

> If you want to schedule a packet sending and let PJON do the job you can use send and call update in the loop although this consumes a little more memory.

@gioblu I think the radio transmission layer use something similar to what you propose, so I think, if it would be not too much resource consuming, then we should give a try to let PJON manage sending packets in background using call from main program loop. Could you prepare the changes? Then I will give it a try and check how much more resources it consumes compare to send_blocking.

gioblu:
@tekka007 @gryzli133  ok I will write down both blocking and non-blocking versions you will just need to paste it in and try :)

gioblu:
Sorry for being late @tekka007 @gryzli133 
To implement the blocking version using `send_packet_blocking`:
1. add at line 23 of `MyTransportPJON.cpp` `#define PJON_RECEIVE_WHILE_SENDING_BLOCKING`
2. to work properly the reception or polling should last some time, for example 1 millisecond, so changing in the function `transportDataAvailable` at line 104 `bus.receive(1000);` (it may have more sense to use a predefined constant / setters and getters for the duration so users can change that). 
3. The `transportSend` function should look like this:
```cpp
bool transportSend(const uint8_t to, const void *data, const uint8_t length, const bool noACK)
{
	const char *datap = static_cast<const char *>(data);
	char *dataToSend = const_cast<char *>(datap);
	uint16_t res = bus.send_packet_blocking(to, dataToSend, length, (noACK) ? (bus.config & ~PJON_ACK_REQ_BIT) : PJON_NO_HEADER);
		switch(res) {
		case PJON_ACK:
			PJON_DEBUG(PSTR("PJON:SND:ACK\n"));
			break;
		case PJON_BUSY:
			PJON_DEBUG(PSTR("!PJON:SND:BUSY\n"));
			break;
		case PJON_FAIL:
			PJON_DEBUG(PSTR("!PJON:SND:FAIL\n"));
			break;
		default:
			PJON_DEBUG(PSTR("!PJON:SND:RSP=%" PRIu8 "\n"), res);
		}
		if (res == PJON_ACK) {
			return true;
		}
	}
	PJON_DEBUG(PSTR("!PJON:SND:TO=%" PRIu8 ",NACK\n"), to);
	return res == PJON_ACK;
}
```
With those changes it should work fine, I also have implemented the `noACK` feature.
The PROs should be less memory footprint and certainty of transmission (the result of transmission can be checked thanks to the blocking operation and the return value). CONS long blocking duration that can break or slow down the user application.

Above @gryzli133 was pointing out the long blocking time in case of disconnection or interference. To avoid this problem you can implement the transport using the PJON's packet buffer and so be non-blocking:
1. implement `transportSend` function accordingly 
```cpp
bool transportSend(const uint8_t to, const void *data, const uint8_t length, const bool noACK)
{
      const char *datap = static_cast<const char *>(data);
      char *dataToSend = const_cast<char *>(datap);
      uint16_t res = bus.send(to, dataToSend, length, (noACK) ? (bus.config & ~PJON_ACK_REQ_BIT) : PJON_NO_HEADER);
      bus.update();
      if(res == PJON_FAIL) { 
        PJON_DEBUG(PSTR("!PJON:SND:FAIL\n"));
        return false;
      } else return true;
}
```
2. to work properly the reception or polling should last some time, for example 1 millisecond, so changing in the function `transportDataAvailable` at line 104 `bus.receive(1000);` (it may have more sense to use a predefined constant / setters and getters for the duration so users can change that). 
3. Add a line to `transportDataAvailable` after line 104, adding `bus.update();`

The main PRO is for sure less time stolen to the application, although, considering that `transportSend` returns before PJON has effectively transmitted the data, it is not possible to know if the packet will be delivered correctly or not, as I have implemented it `transportSend` returns true if there was space in the buffer, and the packet was correctly dispatched, although, the end result of transmission in future may be a failure, so keep it in mind. PJON throws an error in case of transmission failure, but I do not see how to wire it up with the MySensors API, see https://github.com/gioblu/PJON/blob/master/documentation/error-handling.md.

@tekka007 thank you very much for bringing this forward, note that I have not written down some of the required debug prints because I am not really familiar with their format.

gryzli133:
Sorry for being late. I just wanted to check how it is working with the second solution (not blocking) - but forgot to take second Arduino with me. The good news is, even as node without connection to the gateway, the on-board logic is working stable and fast. I will try it soon in real Environment - 30 switches/lights on one node, connected to Domoticz over Gateway. 

For test purpose now without connection to the gateway I have build an multibutton board with max und min cycle time supervision. It is like every 10s one Loop is taking 700ms, every other time there is an avarange of 1-3ms per Loop. This happens only with PJON. Using other MySensors transport I could not see it. I don't think this will disturb the normal use, but maybe it is possible to optimize it. Do have any idea why this can happen? 

The counter is printing min and max loop duration every 1000 loops.

The debug output looks like:
0 (min)
2 (max)
(...) 
1 (min)
2 (max)
37338 TSM:FAIL:RE-INIT
37340 TSM:INIT
38097 TSM:INIT:TSP OK
38099 TSM:INIT:STATID=1
38102 TSF:SID:OK,ID=1
38104 TSM:FPAR
38113 TSF:MSG:SEND,1-1-255-255,s=255,c=3,t=7,pt=0,l=0,sg=0,ft=0,st=OK:
0 (min)
779 (max)
40121 !TSM:FPAR:NO REPLY
40124 TSM:FPAR
40133 TSF:MSG:SEND,1-1-255-255,s=255,c=3,t=7,pt=0,l=0,sg=0,ft=0,st=OK:
1 (min)
20 (max)
42140 !TSM:FPAR:NO REPLY
42142 TSM:FPAR
42151 TSF:MSG:SEND,1-1-255-255,s=255,c=3,t=7,pt=0,l=0,sg=0,ft=0,st=OK:
1 (min)
20 (max)
44160 !TSM:FPAR:NO REPLY
44163 TSM:FPAR
44172 TSF:MSG:SEND,1-1-255-255,s=255,c=3,t=7,pt=0,l=0,sg=0,ft=0,st=OK:
1 (min)
20 (max)
1 (min)
3 (max)
46179 !TSM:FPAR:FAIL
46181 TSM:FAIL:CNT=3
46183 TSM:FAIL:DIS
46184 TSF:TDI:TSL
0 (min)
8 (max)
0 (min)
2 (max)
(...)
0 (min)
2 (max)

The good is, even if the all 12 sensors are trying to send data parallel, the loop still running in 1-30 ms.

1 (min)
3 (max)
523979 !MCO:SND:NODE NOT REG
523986 !MCO:SND:NODE NOT REG
524061 !MCO:SND:NODE NOT REG
524064 !MCO:SND:NODE NOT REG
524067 !MCO:SND:NODE NOT REG
524069 !MCO:SND:NODE NOT REG
524072 !MCO:SND:NODE NOT REG
524076 !MCO:SND:NODE NOT REG
524079 !MCO:SND:NODE NOT REG
524081 !MCO:SND:NODE NOT REG
524084 !MCO:SND:NODE NOT REG
1 (min)
26 (max)

gioblu:
Ciao @gryzli133 thank you very much for your tests and your clear report:
What I do see in the code here:
https://github.com/mysensors/MySensors/blob/46013a5b26d12979df762c1aa3ad7a7eec991eae/hal/transport/PJON/MyTransportPJON.cpp#L45
is that the exponential back-off PJON implements here blocking for the `send_packet_blocking` function: https://github.com/gioblu/PJON/blob/master/src/PJON.h#L671
or here non-blocking (for the `send` and `send_repeatedly` function):
https://github.com/gioblu/PJON/blob/master/src/PJON.h#L922
is missing in `transportSend`. Please comply with the amount of delay between each retry and with the maximum amount of retries, to be coherent with the specification and obtain the best performance. The absence of any back-off may be the culprit.

Another thing I do see here:
https://github.com/mysensors/MySensors/blob/46013a5b26d12979df762c1aa3ad7a7eec991eae/hal/transport/PJON/MyTransportPJON.cpp#L46
This call to `receive` shouldnt be there? There is any reason why is that present in the send function?

The last thing, here:
https://github.com/mysensors/MySensors/blob/46013a5b26d12979df762c1aa3ad7a7eec991eae/hal/transport/PJON/MyTransportPJON.cpp#L102
If the users calls `transportDataAvailable` too seldom, many packets may be lost unheard, remember that SoftwareBitBang uses bit-banging and polling, so it may be required to call `receive` passing a small duration, that could be defined as a macro that can be overridden by the user before the inclusion of `MySensors`. 

I hope one of these small fixes will solve the issue reported by @gryzli133 
Thank you again for your support to the project.

gioblu:
Ah @tekka007 soon we will release PJON v12.1 which will contain new fixes and enhancement. PJON over SoftwareBitBang reached a maximum range between 800 (@ 3.4kB/s or MODE 4) and 2000 meters (@ 1.95kB/s or MODE 1) if a twisted pair is used (ground + data pin) O_O

2000m-Mode-4-Twisted-Pair-8 2kohm-pull-down-60ohm-series-PATCH-2-4-

Here a pic of a  hugely deformed packet going through 2000m wire successfully. In this test it was used one direct connection between Arduino io pins, no amplification or additional hardware.

gryzli133:
Hi Giovanni, 
today I test it in real life test, the results are really good. The only one thing is, that if we send more commends while PJON-Wire is disconnected, they are going to the gateway in strange order after reconnecting. For example, in the test, there is only one possible order to send, every click is changing the LED state (0-1-0-1-0-1 etc), but sometimes the received data after reconnection is in wrong order (0-1-1-0-0-1-0 etc.). This is only important in case of troubles, but maybe there is any easy solution to send the data in First-In -> First-Out order? Because the most important data for one sensor is the last one (f.e. the last state of the Light).

Movie with the test:
https://youtu.be/37ENJcoBgq4

gioblu:
Ciao @gryzli133 I have seen your video, thank you for filming your detailed report :)

I would suggest to @tekka007 to try to address the issues I have underlined in my response above.

If you want to receive packets only once and avoid packet duplication you can use the [packet id feature](https://github.com/gioblu/PJON/blob/master/documentation/configuration.md#packet-identification) here you find an [example](https://github.com/gioblu/PJON/tree/master/examples/ARDUINO/Local/SoftwareBitBang/UsePacketId). Activating this feature PJON will include a packet id field, each packet will carry a unique packet id, PJON takes care that already received packets (with known packet ids) will be filtered out and will not be received more than once, although it will consume a little more memory.

I am available for direct support on gitter if required.
Thank you very much for your support to the project.

gioblu:
@gryzli133 I have watched again your video, you should use at least a pull-down resistor as described in the SoftwareBitBang [README](https://github.com/gioblu/PJON/tree/master/src/strategies/SoftwareBitBang), PJDL is really rugged and it works also if interference is present, although that could cause problems and consume half of the bandwidth available (if it is induced AC).

gioblu:
CIao @tekka007 @gryzli133 indeed PJON is buffering if:
> ...working with the second solution (not blocking)

Sorry, I have missed that, I thought you were still testing https://github.com/mysensors/MySensors/commit/46013a5b26d12979df762c1aa3ad7a7eec991eae

Will look into it.

gioblu:
Ciao @tekka007 today I had a chat on gitter with @gryzli133 
The issue he identified is effectively a non issue. The result of his tests look in this case successful, I have studied the code in its latest state and looks fine. We have broken compatibility in master so I suggest you to stick with 12.1 at least for an initial release, I can then help to migrate to 13.0 afterwards.

Let me know if my help is required.

gioblu:
@gryzli133 about the duration issue you report: that happens because @tekka007 diverted to blocking unbuffered sending (see: https://github.com/mysensors/MySensors/blob/46013a5b26d12979df762c1aa3ad7a7eec991eae/hal/transport/PJON/MyTransportPJON.cpp#L45) so, if a device is not immediately responding, it blocks the loop of the sender up to 3.5 seconds (if the back-off's exponential and limit is implemented as specified). This behavior may mean that sometimes some packets do not go through as easily as others.

gioblu:
@tekka007 I understand your point here in being able to receive but retry as fast as possible to reduce the duration of the function, although that creates a train of packets which saturates the bandwidth and leads to collisions, that may be the cause of the issue reported by @gryzli133 

Here:
https://github.com/mysensors/MySensors/blob/46013a5b26d12979df762c1aa3ad7a7eec991eae/hal/transport/PJON/MyTransportPJON.cpp#L46 
it may have sense to receive for `bus.strategy.back_off(retry)` at least complying with the backoff. You can choose the reduce the maximum amount of retries but that will lead to higher chances of transmission failure. 

Here: 
https://github.com/mysensors/MySensors/blob/46013a5b26d12979df762c1aa3ad7a7eec991eae/hal/transport/PJON/MyTransportPJON.cpp#L104
passing a small duration (in microseconds) to `receive` could help reducing the chances of devices missing packets while doing other stuff.

gryzli133:
@tekka007 would it be possible to lunch beta version MySensors with PJON? I'm currently switching my setup to PJON. Since one week I'm experiencing stable communication for my window contactrons. This week I would like to switch to PJON with some of the light and blinds. But with a house with each Light, Blind, Thermostat, Contactron beeing controlled by MySensors (200+ devices), don't want to go for unofficial release for too long. 

Since one year I'm experiencing 100% stability using mysensors in unusuall configuration -> all devices are connected to multiple gateways (Arduino MEGA). All of them connected to RPi with Domoticz through USB Gateway. But at the moment it starts to be too much USB Hubs in the USB Network, so I want to switch to other wired protocol. PJON looks really promising, so I hope it will be officially supported soon :-)

I'm not the only one using it that way. I have share the code for switches, blinds and contactrons on the biggest Smart Home forum in Poland. I have create classes for the devices, so building arduino with 20 switches is just as less work as change 20 lines of code - so the community using my solution is growing, because of the ease of programming. (Code for blinds: https://github.com/gryzli133/RollerShutterSplit)

Many of the users like the wired solution with USB Gateway as node with lights/blinds, but some of them have problem, that the distance between Arduino and Server is too far for USB - or no cable pulling is possible. Unfortunately using MySensors the way I use it (and many guys in Poland) is just stable as USB Gateway. Using it as Ethernet/MQTT will start hanging in case of Router/Switch/Cable or other network problem. So in the worst case you cannot even switch light. PJON will be here a perfect solution. We could switch to Ethernet/MQTT gateway, and leave gateway as gateway, and use the wired PJON protocol to communicate with nodes that control the light, blinds and other devices. Because as shown on my last video above, there was no problem with node-program hanging on communication problem. So the stability is provided.

tekka007:
@gryzli133 Wow, you have quite some sensors attached :) I'll look into that and review your and @gioblu comments ASAP.

gryzli133:
@tekka007 did you find some time? I have switch 3 arduinos to PJON and it looks really promissing. Using it for blinds, lights and contactrons in my new office since one week with good results. 
image

szurr:
Hi 
How does your setup looks right now?  I meant gateway wires to nodes . Do you use one wire pjon connection to gateway and gateway via USB to server?

gryzli133:
@szurr exactly as you wrote, the nodes (3x arduino mega) are connected through one wire (PJON) to USB-gateway (arduino mega). Gateway is connected through USB port to Domoticz (RPi). At the moment it looks very stable. Next week I want to add next two nodes with contactrons and one node to control my ventilation system. Maybe I will even try to add Ethernet gateway to check the stability in case of network failure, but this should be not a problem, because we have already achieved, that the node works stable independent of the gateway.

szurr:
Great 
Do you have any sensors on gateway?
And whatnis ventilation node? I have right now  test rig on esp to figure needed function but wire node is always better. 
Thanks for your work your blind controller working for my about year now and all "bugs" where eventually my mistakes 😁
Cheers and keep up good work

tekka007:
@gryzli133 Not yet, need to test the arduino library dependency function to have the PJON library added automatically.

paultbarrett:
Been following this conversation and have tried out the _tekka007:PJONSupport_ repo however getting unreliable messages when using the ethernet gateway example included within (W5100) with MYSController or HomeAssistant. Switching to the serial gateway there are no errors and all messages make it through reliably

Network is very small which consists of the following
Arduino Uno (Gateway) -> 20cm cable -> Arduino Nano (Binary Sensor)
_1M resistor across the data line and gnd plus a 160K resistor inline with data line to gateway_

Unreliable messages are seen for all communications including tripped messages, presentations and heartbeats when using ethernet

Debug is turned on in the sketch and can see that PJON is erroring and having to retry 5 times

**Serial console from sensor connected to gateway running ethernet sketch**
```
17298 !MCO:PRO:RC=1
17300 !MCO:PRO:RC=1
17301 !MCO:PRO:RC=1
17303 !MCO:PRO:RC=1
17305 !MCO:PRO:RC=1
17326 PJON:SND:TO=0,LEN=20,RET=1/5
17329 !PJON:SND:FAIL
17349 PJON:SND:TO=0,LEN=20,RET=2/5
17352 !PJON:SND:FAIL
17372 PJON:SND:TO=0,LEN=20,RET=3/5
17375 !PJON:SND:FAIL
17395 PJON:SND:TO=0,LEN=20,RET=4/5
17398 !PJON:SND:FAIL
17419 PJON:SND:TO=0,LEN=20,RET=5/5
17422 !PJON:SND:FAIL
17424 !PJON:SND:TO=0,NACK
17426 !TSF:MSG:SEND,1-1-0-0,s=255,c=3,t=11,pt=0,l=13,sg=0,ft=1,st=NACK:Binary Sensor
17446 PJON:SND:TO=0,LEN=10,RET=1/5
17449 !PJON:SND:FAIL
17464 PJON:SND:TO=0,LEN=10,RET=2/5
17467 PJON:SND:ACK
17469 TSF:MSG:SEND,1-1-0-0,s=255,c=3,t=12,pt=0,l=3,sg=0,ft=2,st=OK:1.0
17585 PJON:SND:TO=0,LEN=7,RET=1/5
17589 !PJON:SND:FAIL
17600 PJON:SND:TO=0,LEN=7,RET=2/5
17603 !PJON:SND:FAIL
17614 PJON:SND:TO=0,LEN=7,RET=3/5
17617 !PJON:SND:FAIL
17629 PJON:SND:TO=0,LEN=7,RET=4/5
17632 !PJON:SND:FAIL
17644 PJON:SND:TO=0,LEN=7,RET=5/5
17647 !PJON:SND:FAIL
17649 !PJON:SND:TO=0,NACK
17651 !TSF:MSG:SEND,1-1-0-0,s=3,c=0,t=0,pt=0,l=0,sg=0,ft=0,st=NACK:
17767 PJON:SND:TO=0,LEN=7,RET=1/5
17770 !PJON:SND:FAIL
17781 PJON:SND:TO=0,LEN=7,RET=2/5
17784 PJON:SND:ACK
17786 TSF:MSG:SEND,1-1-0-0,s=4,c=0,t=0,pt=0,l=0,sg=0,ft=1,st=OK:
```

**Serial console from sensor connected to gateway running serial sketch** 
```
24507 !MCO:PRO:RC=1
24509 !MCO:PRO:RC=1
24510 !MCO:PRO:RC=1
24512 !MCO:PRO:RC=1
24514 !MCO:PRO:RC=1
24516 !MCO:PRO:RC=1
24536 PJON:SND:TO=0,LEN=20,RET=1/5
24539 PJON:SND:ACK
24541 TSF:MSG:SEND,1-1-0-0,s=255,c=3,t=11,pt=0,l=13,sg=0,ft=0,st=OK:Binary Sensor
24561 PJON:SND:TO=0,LEN=10,RET=1/5
24564 PJON:SND:ACK
24565 TSF:MSG:SEND,1-1-0-0,s=255,c=3,t=12,pt=0,l=3,sg=0,ft=0,st=OK:1.0
24681 PJON:SND:TO=0,LEN=7,RET=1/5
24684 PJON:SND:ACK
24686 TSF:MSG:SEND,1-1-0-0,s=3,c=0,t=0,pt=0,l=0,sg=0,ft=0,st=OK:
24801 PJON:SND:TO=0,LEN=7,RET=1/5
24804 PJON:SND:ACK
24806 TSF:MSG:SEND,1-1-0-0,s=4,c=0,t=0,pt=0,l=0,sg=0,ft=0,st=OK:
```

NetworkAnalysis sketch runs with no issues and all looks healthly so does appear to be hardware or connectivity related.

Hope someone might be able to assist as it would be great to use the ethernet gateway

Sketches that are used are the example gateway and a tweaked version of the Binary Sensor

**Gateway**
```
// Enable debug prints to serial monitor
#define MY_DEBUG
#define MY_DEBUG_VERBOSE_PJON

// Enable and select radio type attached
//#define MY_RADIO_RF24
//#define MY_RADIO_NRF5_ESB
//#define MY_RADIO_RFM69
//#define MY_RADIO_RFM95
#define MY_PJON
#define MY_PJON_PIN 7

// Enable gateway ethernet module type
#define MY_GATEWAY_W5100

// W5100 Ethernet module SPI enable (optional if using a shield/module that manages SPI_EN signal)
#define MY_W5100_SPI_EN 4

// Enable Soft SPI for NRF radio (note different radio wiring is required)
// The W5100 ethernet module seems to have a hard time co-operate with
// radio on the same spi bus.
#if !defined(MY_W5100_SPI_EN) && !defined(ARDUINO_ARCH_SAMD)
#define MY_SOFTSPI
#define MY_SOFT_SPI_SCK_PIN 14
#define MY_SOFT_SPI_MISO_PIN 16
#define MY_SOFT_SPI_MOSI_PIN 15
#endif

// When W5100 is connected we have to move CE/CSN pins for NRF radio
#ifndef MY_RF24_CE_PIN
#define MY_RF24_CE_PIN 5
#endif
#ifndef MY_RF24_CS_PIN
#define MY_RF24_CS_PIN 6
#endif

// Enable UDP communication
//#define MY_USE_UDP  // If using UDP you need to set MY_CONTROLLER_IP_ADDRESS or MY_CONTROLLER_URL_ADDRESS below

// Enable MY_IP_ADDRESS here if you want a static ip address (no DHCP)
#define MY_IP_ADDRESS 10,100,60,252

// If using static ip you can define Gateway and Subnet address as well
#define MY_IP_GATEWAY_ADDRESS 10,100,60,1
#define MY_IP_SUBNET_ADDRESS 255,255,255,0

// Renewal period if using DHCP
//#define MY_IP_RENEWAL_INTERVAL 60000

// The port to keep open on node server mode / or port to contact in client mode
#define MY_PORT 5003

// Controller ip address. Enables client mode (default is "server" mode).
// Also enable this if MY_USE_UDP is used and you want sensor data sent somewhere.
//#define MY_CONTROLLER_IP_ADDRESS 192, 168, 178, 254
//#define MY_CONTROLLER_URL_ADDRESS "my.controller.org"

// The MAC address can be anything you want but should be unique on your network.
// Newer boards have a MAC address printed on the underside of the PCB, which you can (optionally) use.
// Note that most of the Arduino examples use  "DEAD BEEF FEED" for the MAC address.
#define MY_MAC_ADDRESS 0xA0, 0xF3, 0xBE, 0x25, 0xF4, 0xE3


// Enable inclusion mode
#define MY_INCLUSION_MODE_FEATURE
// Enable Inclusion mode button on gateway
//#define MY_INCLUSION_BUTTON_FEATURE
// Set inclusion mode duration (in seconds)
#define MY_INCLUSION_MODE_DURATION 60
// Digital pin used for inclusion mode button
//#define MY_INCLUSION_MODE_BUTTON_PIN  3

// Set blinking period
#define MY_DEFAULT_LED_BLINK_PERIOD 300

// Flash leds on rx/tx/err
// Uncomment to override default HW configurations
//#define MY_DEFAULT_ERR_LED_PIN 7  // Error led pin
//#define MY_DEFAULT_RX_LED_PIN  8  // Receive led pin
//#define MY_DEFAULT_TX_LED_PIN  9  // Transmit led pin

#if defined(MY_USE_UDP)
#include <EthernetUdp.h>
#endif
#include <Ethernet.h>
#include <MySensors.h>

void setup()
{
	// Setup locally attached sensors
}

void presentation()
{
	// Present locally attached sensors here
}

void loop()
{
	// Send locally attached sensors data here
}
```


**Sensor**
```
// Enable debug prints to serial monitor
#define MY_DEBUG
#define MY_DEBUG_VERBOSE_PJON

#define MY_PJON
#define MY_PJON_PIN 7

#include <MySensors.h>
#include <Bounce2.h>

Bounce debouncer1 = Bounce(); 
int oldValue1=-1;

Bounce debouncer2 = Bounce(); 
int oldValue2=-1;

#define SKETCH_NAME "Binary Sensor"
#define SKETCH_MAJOR_VER "1"
#define SKETCH_MINOR_VER "0"

#define PRIMARY_CHILD_ID 3
#define SECONDARY_CHILD_ID 4

#define PRIMARY_BUTTON_PIN 2   // Arduino Digital I/O pin for button/reed switch
#define SECONDARY_BUTTON_PIN 3 // Arduino Digital I/O pin for button/reed switch

// Change to V_LIGHT if you use S_LIGHT in presentation below
MyMessage msg1(PRIMARY_CHILD_ID, V_TRIPPED);
MyMessage msg2(SECONDARY_CHILD_ID, V_TRIPPED);

void setup()
{
	// Setup the buttons
	pinMode(PRIMARY_BUTTON_PIN, INPUT_PULLUP);
	pinMode(SECONDARY_BUTTON_PIN, INPUT_PULLUP);

  // After setting up the button, setup debouncer
  debouncer1.attach(PRIMARY_BUTTON_PIN);
  debouncer1.interval(5);

  debouncer2.attach(SECONDARY_BUTTON_PIN);
  debouncer2.interval(5);
  
}

void presentation()
{
	// Send the sketch version information to the gateway and Controller
	sendSketchInfo(SKETCH_NAME, SKETCH_MAJOR_VER "." SKETCH_MINOR_VER);
  delay(100);
	// Register binary input sensor to sensor_node (they will be created as child devices)
	// You can use S_DOOR, S_MOTION or S_LIGHT here depending on your usage.
	// If S_LIGHT is used, remember to update variable type you send in. See "msg" above.
	present(PRIMARY_CHILD_ID, S_DOOR);
  delay(100);
	present(SECONDARY_CHILD_ID, S_DOOR);
  delay(100);
}

// Loop will iterate on changes on the BUTTON_PINs
void loop()
{

  debouncer1.update();
  // Get the update value
  int value1 = debouncer1.read();

  if (value1 != oldValue1) {
     // Send in the new value
     send(msg1.set(value1==HIGH ? 1 : 0));
     oldValue1 = value1;
  }

  debouncer2.update();
  // Get the update value
  int value2 = debouncer2.read();

  if (value2 != oldValue2) {
     // Send in the new value
     send(msg2.set(value2==HIGH ? 1 : 0));
     oldValue2 = value2;
  }
}

```

gryzli133:
@paultbarrett nice to see someone new is interested in using PJON together with MySensors. 

MySensors with EthernetShild is somehow "pausing" the main program (loop). I saw this as a issue trying to use it in standalone gateway as node solution. It is working well as USB-Gateway, but hanging as LAN-Gateway. I think you are facing the same problem. So everytime when the Gateway is communicating over LAN, then PJON and other code are paused.

The biggest problem comes, when the gateway has no LAN connection and then it is hanging in send/receive process, causing all other logic is hanging. The good is, when using not blocking PJON protocol, the node will still work reliable, all local sensors (f.e. light switch and relay) are working well, while the Gateway is hanging and ignoring PJON.

gioblu:
Ciao @paultbarrett @gryzli133 thank you for your feedback.
I have the sensation the problem may be caused by the software reading the LAN side, while SoftwareBitBang packets are lost unheard on the other. 

The back-off fixes the issue at the cost of wasted bandwidth in the SoftwareBitBang bus. I do see that the amount of retries and the delay in between them do not comply with the specification. Complying with the specification will likely partially solve the issue (no error messages, but still a lot of retries). 

To reduce the amount of retries you should poll the SoftwareBitBang bus for a longer time and so reduce the chances of loosing a packet while handling the LAN side. In the actual PJON master we have added a minimum polling time for each strategy: https://github.com/gioblu/PJON/blob/master/src/strategies/SoftwareBitBang/SoftwareBitBang.h#L44
this duration is then used by higher level routing classes to understand if they need to poll that strategy and for how much time. Buffered strategies like Ethernet or Serial have set a value of 0, strategies based on polling have a value of 1 millisecond. Testing master now I can confirm a PJON router that uses Ethernet on one side and SoftwareBitBang on the other transmits the packet in the worst case twice (on the SoftwareBitBang side). 

That is why some messages above I was suggesting to set a slightly higher polling time in the transport.
If any help on the implementation is required I am here

paultbarrett:
@gioblu  @gryzli133  thanks for the responses.

Whilst I can write my own sketches my knowledge is not enough to be able to implement the change in the library to poll the PJON bus for a longer period of time. Can you provide any pointers of where i am to start.

gioblu:
Sure @paultbarrett 

Going up in the issue the last referenced commit looks: [69a0b43](https://github.com/mysensors/MySensors/commit/69a0b430ad6fa5aa0e5f4044d638f597362161c3) 

https://github.com/mysensors/MySensors/commit/69a0b430ad6fa5aa0e5f4044d638f597362161c3#diff-f93d9576c0213eaef64d26e97cc5139aR102
`transportDataAvailable` should call `bus.receive()` passing a parameter with a value of 1000, that would force reception for at least 1 millisecond

paultbarrett:
@gioblu , thanks for the tip.
I tested out 1ms however still had some random errors where the 5 retries were exceeded.
I increased the value to 2ms and no more timeouts. Retries never exceeds 2/5

Further testing needs to be done but looking good so far.

Will come back to advise of the results after some days of testing so that this may be incorporated into the library if needed

gioblu:
Ciao @paultbarrett thank you for your feedback, happy to hear now works properly.
@tekka007 @gryzli133 if the non-blocking version of the transport works fine as well, it may has sense to use the non-blocking version, that would comply with the back-off specification and be more reliable at the cost of some ram and program memory.

gryzli133:
Ciao @gioblu after long time testing now, I can confirm, that the protocol in non-blocking version is very stable and reliable. I was really trying to find some bugs, but I could not ;-)

gioblu:
Ciao @gryzli133 that's a good news :), so @tekka007 have only to change the code using the non-blocking version and remember to poll for at least 2 milliseconds, then the transport should be ready to go.

Carterkrieg:
Hallo to all Mysensors and PJON gurus, Especially @tekka007  and @gioblu !
Your work is really appreciated!I Just want to confirm stability and reliability (although just few simple nodes). 

Just quick question: Is Repeater function usable with PJON? Because when I set MY_PARENT_NODE_ID (with or without PARENT_NODE_IS_STATIC), the GW message from NODE 210 (simple dallas s.) is always : 0;255;3;0;9;1158230 TSF:MSG:READ,210-210-0,s=0,c=1,t=0,pt=7,l=5,sg=0:21.5
210;0;1;0;0;21.5... 
although, 210 is connected through (from pin 7 on 210 to p7 on 230 to GW p7) Node 230 (ofc  with repeater enabled +N210 has defined MY_PARENT_NODE_ID (230)).

Or am I reading the LAST NODE part in msg incorrectly? 

Thanks and keep up the good work.

gryzli133:
Hi @tekka007, Hi @gioblu,

how big are the chances that PJON will be available soon in the official version of MySensors? I'm testing it now for a couple of months and facing no problems at all. Super stable and easy expandable network (just one additional cable to the sensor). I moved about 20% of the devices to PJON (Office), but the target is to go 100% with PJON (house and garage). I would like to migrate another devices first with official version, to be sure about the compatibility in the future.

Anyway thank you for your support until now! ;-)

tekka007:
Hi @gryzli133. Thank you for the feedback on the PJON layer stability. We all have limited bandwith at the moment (and hence the delay with this and other PRs). Nevertheless, we'll consider this PR for the next MySensors release. If things go well, ETA: XMASish... 🤞

tekka007:
Hi @gryzli133 & @gioblu - I've updated this PR to include PJON 13.0. If time allows, can you retest and provide feeback?

paultbarrett:
I have retested this in my small environment and it is working as expected.

________________________________

gryzli133:
Hi @tekka007, can you send me download link to the latest version? Or maybe even better - can you tell me how can I find it by myself next time?
Did you also include the changes for using the non-blocking version and polling for at least 2 milliseconds?

tekka007:
@gryzli133 https://github.com/tekka007/MySensors/tree/PJONSupport

No, I didn't inlcude the changes you mentioned - please feel free to add them

gryzli133:
> Ciao @gryzli133 that's a good news :), so @tekka007 have only to change the code using the non-blocking version and remember to poll for at least 2 milliseconds, then the transport should be ready to go.

Hi @gioblu, could you take care about that two changes to be included in the latest release? Then I will start testing new version. Thank you in advance!

gioblu:
@gryzli133 where I can find your source?

gryzli133:
@gioblu at the moment I'm still using the old version 98d3d33 with the modifications you gave me for non-blocking version:

> 1. implement `transportSend` function accordingly
> 
> ```c++
> bool transportSend(const uint8_t to, const void *data, const uint8_t length, const bool noACK)
> {
>       const char *datap = static_cast<const char *>(data);
>       char *dataToSend = const_cast<char *>(datap);
>       uint16_t res = bus.send(to, dataToSend, length, (noACK) ? (bus.config & ~PJON_ACK_REQ_BIT) : PJON_NO_HEADER);
>       bus.update();
>       if(res == PJON_FAIL) { 
>         PJON_DEBUG(PSTR("!PJON:SND:FAIL\n"));
>         return false;
>       } else return true;
> }
> ```
> 
> 2. to work properly the reception or polling should last some time, for example 1 millisecond, so changing in the function `transportDataAvailable` at line 104 `bus.receive(1000);` (it may have more sense to use a predefined constant / setters and getters for the duration so users can change that).
> 3. Add a line to `transportDataAvailable` after line 104, adding `bus.update();`

gioblu:
@tekka007 I have done as  @gryzli133 requested, here it is https://github.com/tekka007/MySensors/pull/1/files

gryzli133:
@tekka007 I just wanted to make some test with new version of MySensors, but I'm getting some troubles for ESP8266. I have copied my portable version of Arduino 1.8.12 where the first PJON implementation is working fine. Everything is the same, just MySensors is updated. Now with cdc0ef2 I'm getting this faults:
image

I will continue tests with serial and Ethernet gateway, but for fast setup wifi gateway is very useful.

tekka007:
Hi @gryzli133, did you update the esp8266 board files to the latest version, i.e. 2.7.4?

gryzli133:
Thank you @tekka07, I did some tests and it looks very good so far ;-) Do you already have a plan when will be next official release and if PJON will be included? Will PJON be also supporte by Multitransport feature?

gioblu:
Ciao @tekka007 any update on this? When the new version will be released?

dgertych-monterail:
Great to see PJON support merged 🎆

kofec:
Can we have example for RS485 and PJON ?

fabyte:
@tekka007  @gioblu  So the license issue has been solved? I wonder how that went?

tekka007:
@fabyte In dev for testing & feedback. Evaluating setup using the Arduino library dependency manager (https://arduino.github.io/arduino-cli/latest/library-specification/) to link to PJON

gryzli133:
Hi @tekka007, 

do you know when MySensors will be released with official PJON support? At the moment I'm currently running a MySensors network connected with PJON including:
- 1x Gateway - Arduino Mega with Ethernet Shield
- 18x Thermostat Nodes - ESP8266 with switch off WiFi (used only for OTA)
- 21x Arduino Mega - controlling lights, blinds and contactrons

At the moment I'm facing no problem, so I would say - PJON integration is ready for community use ;-)

Many thanks for your support @tekka007 and @gioblu !

PR #1290, opened 2019-05-21
MySensorsCore: new sendAsIs() function

pwieser:
Do not force a C_SET command, letting the client use a C_INTERNAL one,
e.g. to send the library version to the controller.

### DIFF
--- a/core/MySensorsCore.cpp
+++ b/core/MySensorsCore.cpp
@@ -322,8 +322,13 @@ bool _sendRoute(MyMessage &message)
 
 bool send(MyMessage &message, const bool enableAck)
 {
-	message.sender = getNodeId();
 	mSetCommand(message, C_SET);
+	return sendAsIs(message, enableAck)
+}
+
+bool sendAsIs(MyMessage &message, const bool enableAck)
+{
+	message.sender = getNodeId();
 	mSetRequestAck(message, enableAck);
 
 #if defined(MY_REGISTRATION_FEATURE) && !defined(MY_GATEWAY_FEATURE)
--- a/core/MySensorsCore.h
+++ b/core/MySensorsCore.h
@@ -162,6 +162,7 @@ bool sendSketchInfo(const __FlashStringHelper *name, const __FlashStringHelper *
 * @return true Returns true if message reached the first stop on its way to destination.
 */
 bool send(MyMessage &msg, const bool ack = false);
+bool sendAsIs(MyMessage &msg, const bool ack = false);
 
 /**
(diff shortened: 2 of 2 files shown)


scalz:
Hello,

If you need to get the MySensors library version, I_INTERNAL messages are usually used by core functions, not directly exposed to the user like this, I think.  
So I'm not sure if a sendAsIs-like user function is really needed. I let others answer :)

What about something like this? I've not tried but it might also be more user friendly in user sketch

```
bool sendMySensorsVersion(const bool ack)
{
	return _sendRoute(build(_msgTmp, GATEWAY_ADDRESS, NODE_SENSOR_ID, C_INTERNAL, I_VERSION,
		                           ack).set(MYSENSORS_LIBRARY_VERSION));

}
```
Do you know that you can keep track of your firmware revisions with sendSketchInfo?
Here more infos about MySensors Api (battery level, sketchInfo etc): https://www.mysensors.org/apidocs-beta/group__MySensorsCoregrp.html

tekka007:
This functionality already [exists](https://github.com/mysensors/MySensors/blob/master/core/MySensorsCore.cpp#L353-L358)- simply call:

`present(NODE_SENSOR_ID, S_ARDUINO_NODE)
`

In addition, the controller can request the library version by sending a `C_INTERNAL | I_PRESENTATION` message to the node.

pwieser:
Hello Olivier, 

I do not understand your point : the present() function publishes the sensor id and type, not the library version. 
Mine is to let the node advertise the controller of the library version it has been compiled with. 
Or am I missing something ? 

Regards 
Pierre

pwieser:
Hello Scalz, 

Your proposal would be OK for my own immediate need. 
I just tried to patch the code a minima, while keeping a generally usable method, i.e. for other possible use cases... 

More, in my understanding, sendSketchInfo() sends name and version of the sketch, not of the library version it has been compiled with. 
So, yes, one could use the "version" part of the function as the library version instead of the sketch version, but I think it would be more "user-friendly" to be able to publis the two informations. 

Regards 
Pierre

tekka007:
> Hello Olivier, I do not understand your point : the present() function publishes the sensor id and type, not the library version. 

This is exactly what `present()` does when called with `NODE_SENSOR_ID` as parameter (see [here](https://github.com/mysensors/MySensors/blob/master/core/MySensorsCore.cpp#L353-L358)). Btw, the library version is automatically sent during bootup using this exact command, see [here](https://github.com/mysensors/MySensors/blob/master/core/MySensorsCore.cpp#L232) for reference.

Issue #1272, opened 2019-03-24
Rename RadioSettingGrpPub to TransportSettingGrpPub


mfalkvidd:
https://www.mysensors.org/apidocs/group__RadioSettingGrpPub.html has the headline "Radio selection". In MySensors lingo, we probably mean "Transport selection", especially since RS485 is included but does not use radio.

If we rename the group and the text, it might become easier for users to understand which different transports are supported by MySensors.

I am not sure why the SPI stuff is in this group though. Is there someplace else where it would fit better? If it does fit well within the transport section, could we elaborate on why?

PR #1295, opened 2019-05-30
Rename RadioSettingGrpPub to TransportSettingGrpPub

mfalkvidd:
https://www.mysensors.org/apidocs/group__RadioSettingGrpPub.html
had the headline "Radio selection". In MySensors lingo, we
probably mean "Transport selection", especially since RS485 is
included but does not use radio.

By renaming the group and the text, it might become easier for
users to understand which different transports are supported
by MySensors.

This fixes https://github.com/mysensors/MySensors/issues/1272

### DIFF
--- a/MyConfig.h
+++ b/MyConfig.h
@@ -192,16 +192,16 @@
 
 
 /**
- * @defgroup RadioSettingGrpPub Radio selection
+ * @defgroup TransportSettingGrpPub Transport selection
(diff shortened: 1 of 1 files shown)


mfalkvidd:
Updated doxygen page: https://ci.mysensors.org/job/MySensors/job/MySensors/job/PR-1295/1/Doxygen_20HTML/

PR #1308, opened 2019-07-13
Fix issues after CPPcheck 1.88 update

tekka007:
CI:
- Cppcheck updated to 1.88: added compatibility + performance checks, disabled information messages, fixed newly reported code issues
- astyle updated to 3.1
- updated bootstrap-dev.sh version checks


examples/SecretKnockSensor/SecretKnockSensor.ino
```diff
@@ -182,7 +182,7 @@ void listenToSecretKnock()
 
 	int currentKnockNumber = 0;               // Position counter for the array.
 	int startTime = millis();                 // Reference for when this knock started.
-	int now = millis();
+	int now;
```
mfalkvidd: Not related to your changes, but assigning the return value of millis() to a signed integer will only work the first 32 seconds after boot. Better use unsigned long or uint32_t.

hal/transport/RFM69/driver/old/RFM69_old.cpp
```diff
@@ -58,7 +58,7 @@ bool RFM69::initialize(uint8_t freqBand, uint8_t nodeID, uint8_t networkID)
 
 		/* 0x07 */ { REG_FRFMSB, (uint8_t) (freqBand==RFM69_315MHZ ? RF_FRFMSB_315 : (freqBand==RFM69_433MHZ ? RF_FRFMSB_433 : (freqBand==RFM69_868MHZ ? RF_FRFMSB_868 : RF_FRFMSB_915))) },
 		/* 0x08 */ { REG_FRFMID, (uint8_t) (freqBand==RFM69_315MHZ ? RF_FRFMID_315 : (freqBand==RFM69_433MHZ ? RF_FRFMID_433 : (freqBand==RFM69_868MHZ ? RF_FRFMID_868 : RF_FRFMID_915))) },
-		/* 0x09 */ { REG_FRFLSB, (uint8_t) (freqBand==RFM69_315MHZ ? RF_FRFLSB_315 : (freqBand==RFM69_433MHZ ? RF_FRFLSB_433 : (freqBand==RFM69_868MHZ ? RF_FRFLSB_868 : RF_FRFLSB_915))) },
+		/* 0x09 */ { REG_FRFLSB, (uint8_t) 0x00 /*duplicateValueTernary: (freqBand==RFM69_315MHZ ? RF_FRFLSB_315 : (freqBand==RFM69_433MHZ ? RF_FRFLSB_433 : (freqBand==RFM69_868MHZ ? RF_FRFLSB_868 : RF_FRFLSB_915)))*/ },
```
mfalkvidd: I would prefer to suppress the cppcheck message here. If, in the future, any of the RF_FRFLSB_* macros are changed to something else than 0x00, we would want those changes to have an effect.

PR #1309, opened 2019-07-14
Code maintenance

### LINKED ISSUE #1222: Consolidate message access
MySensors currently provides two ways to check if a message is an acknowledgement/echo:  _mGetAck(message)_ and _message.isAck()_. Is there a reason for providing two ways to do the same thing? Could we clean up the API by removing one of them?

There are other instances of the same duality, for example [mGetCommand](https://www.mysensors.org/apidocs/group__MyMessagegrp.html#gae2370fc35a3e46b0d83d1efeb9947701) and [message.getCommand()](https://www.mysensors.org/apidocs/classMyMessage.html#a0c2e46ac36c11e7c2e13513b6bd76327) so when updating we should be as consistent as possible.

tekka007:
- Define getters/setters for MyMessage class, fixes #1222
- Centralize payload length verification
- Fix doxygen tags
- add _process() recursive call check


mfalkvidd:
With this change, MockMySensors uses 328 bytes more flash, for the same functionality. That's an increase by 2,3% (from 13,990 bytes to 14,318).

We're already tight on space on atmega328 for many use cases.

Is there a way to make these changes without using so much extra flash? If not, what do we recommend users who upgrade and no longer can fit their sketches?

tekka007:
Using Arduino IDE 1.8.9 and AVR board defs 1.6.23 I :
MockMySensors.ino / Arduino UNO

with current dev (latest commit [36ec4f8](https://github.com/mysensors/MySensors/commit/36ec4f862498e585a660ec7f5ba1ff573c659c5b)):
```
Sketch uses 14296 bytes (44%) of program storage space. Maximum is 32256 bytes.
Global variables use 835 bytes (40%) of dynamic memory, leaving 1213 bytes for local variables.
```
with this PR:
```
Sketch uses 14318 bytes (44%) of program storage space. Maximum is 32256 bytes.
Global variables use 835 bytes (40%) of dynamic memory, leaving 1213 bytes for local variables. 
```
How did you get your numbers?

mfalkvidd:
Interesting. I used latest development (36ec4f86) and your branch (3bf23c00b81), respectively. Arduino Pro Mini.

But when I re-run the compilation now, I get this for development
```
Sketch uses 14294 bytes (46%) of program storage space. Maximum is 30720 bytes.
Global variables use 835 bytes (40%) of dynamic memory, leaving 1213 bytes for local variables. Maximum is 2048 bytes.
```

and this for your branch:
```
Sketch uses 14316 bytes (46%) of program storage space. Maximum is 30720 bytes.
Global variables use 835 bytes (40%) of dynamic memory, leaving 1213 bytes for local variables. Maximum is 2048 bytes.
```

My guess is that the Arduino IDE reused some cached code. Just disregard my earlier comment. +22 bytes is fine.

tekka007:
> Just disregard my earlier comment. +22 bytes is fine.

Just to be clear: this is a maintenance PR that adds - as the title implies - no additional functionality. The important changes harden the MyMessage class and its use as a core component of the framework. In that respect, even code size changes of 100+ bytes are acceptable if they increase the overall reliability of the framework.

PR #1317, opened 2019-07-22
Move noACK control to transportSendWrite()

tekka007:
- Handle noACK flags globally: BC messages & passive mode
- Follow up of #1316


mfalkvidd:
Could you expand on the reason for this change? Without understanding the intention, it is hard to give feedback on the PR.

How come the nrf drivers aren't affected?

tekka007:
@mfalkvidd I've updated the description: This PR centralizes the noACK handling, RF24 was updated previously.

PR #1328, opened 2019-07-28
Fix MySigningAtsha204Soft bug in Gateway mode 

YuanWujin:
I use one ESP8266 module as the mysensors MQTT gateway,and another ESP8266 module as a mysensors client device.

This gateway does not use MY_SIGNING_NODE_WHITELISTING.

I  connect the mysensors gateway to the Domoticz server,and setup the client devices UI succesfully. 

When I send a msg from client to gateway, the gateway works OK.
But when I tried to control Domoticz UI to change mysensors client devices status,the ESP8266 mysensors client encountered a error :
```
!SGN:VER:FAIL
!TSF:MSG:SIGN VERIFY FAIL
```
The version of Mysensors I used is 2.3.1 .
I modified the source code to fix this bug by adding the Macro MY_SIGNING_NODE_WHITELISTING to hide DO_WHITELIST() )


tekka007:
included in #1347

PR #1336, opened 2019-08-10
We fix an infinite loop if Power-down mode enabled

mysensors-rus:
If the bootloader puts the SPI flash into Power-down mode, then thanks to busy () we have an infinite loop and it doesn’t get to sending the SPIFLASH_WAKE command.


tekka007:
included in #1347

PR #1338, opened 2019-08-13
Add log parser

mfalkvidd:
Add the log parser to git so we have version control/history, can create issues and can receive pull requests.


tekka007:
Can this file be loaded from the log parser page?

mfalkvidd:
No. Github prevents loading javascript remotely, by setting text/plain as content type.
There are some workarounds, but they all seem to be cludgy.
https://stackoverflow.com/a/19285701

mfalkvidd:
But maybe Jenkins kan push the file to mysensors.org, just like it does with doxygen?

henrikekblad:
I already do some magic parsing of the articles to handle the inclusion (and caching) of code from github. Might be possible to adopt it to allow inclusion of the logparser code.

mfalkvidd:
Don't merge yet. Hek proposed some changes through slack.

mfalkvidd:
Ready for merge. Can now be run offline by opening logparser.html in a web browser.

PR #1339, opened 2019-08-13
Add id failed to logparser

mfalkvidd:
Add message for ID:FAIL to log parser.

Before:
logparser_before

After:
logparser_after


logparser.js
```diff
@@ -256,6 +256,7 @@ var match = [
 { re: "TSM:ID", d: "Transition to <b>Request Id</b> state" },
 { re: "TSM:ID:OK,ID=(\\d+)", d: "Node id <b>$1</b> is valid" },
 { re: "TSM:ID:REQ", d: "Request node id from controller" },
+{ re: "!TSM:ID:FAIL", d: "Did not receive an ID from controller. Is your controller correctly configured?" },
```
Yveaux: TSM:ID:REQ", d: "Request **node id** from controller
How about:
!TSM:ID:FAIL", d: "Did not receive **a node id** from controller. Is your controller correctly configured?"

PR #1353, opened 2019-10-10
RPI: Add support for RPI4 and NRF24

phil2fer:
Sorry there are a build Warning but it works.
if someone would help me to solve this
The binary works as expected on the RPI4 with NRF24 radio

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ GATEWAY_OBJECTS=$(patsubst %.c,$(BUILDDIR)/%.o,$(GATEWAY_C_SOURCES)) $(patsubst 
 
 INCLUDES=-I. -I./core -I./hal/architecture/Linux/drivers/core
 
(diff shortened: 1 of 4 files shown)


d-diot:
Hi, thanks for this commit!
I have tested it with a Rpi4 and a NRF24 and RFM69 radio modules. The code compiles and the gateways starts, but for both radio modules i got this error:

DEBUG !TSM:INIT:TSP FAIL


More details here
https://forum.mysensors.org/topic/4803/building-a-raspberry-pi-gateway/1036

phil2fer:
try to use a default /boot/config.txt (without your custom configuration of GPIO port)
and compile mysensor gateway with this option :
`--my-transport=rf24 --my-gateway=ethernet --my-port=5003`

tekka007:
see #1364

PR #1358, opened 2019-10-31
Add MY_RS485_DE_INVERSE define to invert the device enable pin polarity

ltigges:
RS485 ICs like the SN65HVD230 (CAN Transceiver) have an low-active device enable. 

This has been tested with an ATMEGA 328p and an SN65HVD230

PR #1360, opened 2019-11-04
Make it clear when checks pass

mfalkvidd:
#1358 (and others) showed that the butler check messages can be made clearer.

Hopefully, this change makes it more clear when there are no problems.

### DIFF
--- a/.ci/butler.sh
+++ b/.ci/butler.sh
@@ -4,15 +4,15 @@ cd ..
 
 result=0
 
-echo "No subjects are of invalid size<br>" > too_long_subjects.txt
-echo "No subjects with leading lower case characters<br>" > leading_lowercases.txt
-echo "No subjects with trailing periods<br>" > trailing_periods.txt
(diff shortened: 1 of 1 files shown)


mkaiser:
Continueing the CI discussion from #1358

If a build fails due to an astyle issue -  where is the suggested restyling.patch located? Also I did not receive any mails...

```
I am afraid your coding style is not entirely in line with the MySensors prefered style.
A mail with a patch has been sent to you that, if applied to your PR, will make it follow the MySensors coding standards.
You can apply the patch using:
git apply restyling.patch
```

mfalkvidd:
@fallberg was able to dig out the patch for me. I don't know how to find it unfortunately.

mfalkvidd:
Historically, we've had a few (maybe 3-5) issues with people not getting email. It works for most people, but sometimes it doesn't. So far we have not been able to figure out what causes the problem to happen.

PR #1370, opened 2019-11-15
ESP8266: Update to 2.6.1

tekka007:
changes, see https://github.com/esp8266/Arduino/releases


mfalkvidd:
@tekka007 do you know if this affects compatibility with older versions of the esp8266 board definitions? (I e should we not any compatibility requirements in the release notes?)

tekka007:
@mfalkvidd Yes it does, users will have to upgrade ESP8266 board defs to 2.6.1 for MySensors 2.3.2 release. I will provide a list of all board def versions this release is tested with.

PR #1378, opened 2020-01-05
Support also development on OpenSuse

mvdw:
Support also OpenSuse. On OpenSuse Leap 15.1, the OSTYPE is simply 'linux'


mvdw:
This request  is only for https://github.com/mysensors/MySensors/pull/1378/commits/1424148d16e7d76229573c471e6efd18405fce44
I did something wrong I guess...

Also the commit message is not OK (trailing dot) due to:
"The '.git/hooks/commit-msg' hook was ignored because it's not set as executable."

mfalkvidd (approved):

PR #1380, opened 2020-01-06
Fix logparser frequency regexp

henrikekblad:


mfalkvidd:
Verified, works:
image

PR #1401, opened 2020-03-30
Fix compile errors for Adafruit Feather M0

DCEM:
I would like to be able to use Adafruit Feahter M0 Hardware.
This should actually fix it for multiple boards.

I do not know why SERIAL_PORT_USBVIRTUAL and
SERIAL_PORT_MONITOR not defined here:
...ysensors/ArduinoHwSAMD/blob/master/variants/mysensors_gw/variant.h
I also do not understand the difference between MY_SERIALDEVICE and
MY_DEBUGDEVICE since My MY_DEBUGDEVICE semmst to be MY_SERIALDEVICE
for all architectures:
.../mysensors/MySensors/tree/development/hal/architecture

The way I understand the "Arduino way" to do it would have been to
use SERIAL_PORT_MONITOR instead of MY_DEBUGDEVICE
or alternatively use "#define MY_DEBUGDEVICE SERIAL_PORT_MONITOR"

Because I do not own a "MySensors SAMD Board" I cannot test the
outcome I do not want to touch the:
...ysensors/ArduinoHwSAMD/blob/master/variants/mysensors_gw/variant.h

Therfore i suggest the following solution:

#ifndef MY_SERIALDEVICE
 #ifdef SERIAL_PORT_MONITOR
  #define MY_SERIALDEVICE SERIAL_PORT_MONITOR
 #else
  #define MY_SERIALDEVICE SerialUSB
 #endif
#endif


DCEM:
Hi,  I'm sorry I'm not experienced in these matters and I'm proberbly doing it wrong.

The errormesagge is:
Body lines that are too wide (>72 characters):
The way I understand the "Arduino way" to do it would have been to
or alternatively use "#define MY_DEBUGDEVICE SERIAL_PORT_MONITOR"
Because I do not own a "MySensors SAMD Board" I cannot test the

The longest line seems to be 67 charakters.

Sorry to bother you with this, but what is the issue here?

Thank you
DCEM

PR #1414, opened 2020-05-01
Multitransport implementation

tekka007:
MySensors multitransport implementation - WIP
- supports multiple transports with RX queue
- RFM95 radio can run in RFM69 compatibility mode
- RFM69 de-duplication
- centralized encryption
- other updates and improvements, tbd


tekka007:
Closing after inactivity, use case needs to be re-evaluated

PR #1418, opened 2020-05-13
Add sleep handler to optimize low-power sleep

SinishaDjukic:
Sleep handler is invoked right before entering the sleep function, as well a just after the sleep completes. Applications can use this handler to turn off peripherals or put pins into a best possible low power state according to the concrete hardware design.


tekka007:
The current AVR GCC is v7.3.0 (Arduino Boards v1.8.2) - you can use a weak function definition instead of MY_SLEEP_HANDLER.

SinishaDjukic:
> The current AVR GCC is v7.3.0 (Arduino Boards v1.8.2) - you can use a weak function definition instead of MY_SLEEP_HANDLER.

I followed the same approach as MY_INDICATION_HANDLER, which is somewhat more explicitly visible for the developers, but if weak functions are preferred now I can revise.

tekka007:
I totally agree, please go ahead with your implementation. We may reconsider that in a later update.

SinishaDjukic:
Ready for review & merge

PR #1425, opened 2020-05-31
Added a warning about incorrect flags being used

Avamander:
Made the preprocessor throw an error if an ancient define is specified, user must update their defines to avoid very difficult to diagnose issues further on. 

I personally stumbled upon it. I hope this PR avoids the headache for someone else.

### DIFF
--- a/MyConfig.h
+++ b/MyConfig.h
@@ -420,6 +420,16 @@
 #define MY_RF24_CHANNEL (76)
 #endif
 
+// legacy - remove for 3.0.0
+/**
+* @def RF24_DATARATE
+* @brief Define this to specify a specific wireless data rate used by the RF24
+* @deprecated This flag is deprecated and replaced by @ref MY_RF24_DATARATE
+*/
+#ifdef RF24_DATARATE
+#error RF24_DATARATE is deprecated, use MY_RF24_DATARATE instead!
+#endif
+
 /**
  * @def MY_RF24_DATARATE
  * @brief RF24 data rate.


mfalkvidd:
I did some digging and the names were changed in the release of 2.0.0 (1.5.4 which was the last 1.x release had the old variants without MY_)

The commit was f8ee37623074086767715d92bfb8ff2ada8e5079

In the same commit, these were also updated so maybe these should have deprecation warnings as well:
```
RF24_CHANNEL -> MY_RF24_CHANNEL
RF24_BASE_RADIO_ID -> MY_RF24_BASE_RADIO_ID
RF24_CE_PIN -> MY_RF24_CE_PIN
RF24_CS_PIN -> MY_RF24_CS_PIN
RF24_PA_LEVEL -> MY_RF24_PA_LEVEL
RF24_PA_LEVEL_GW -> not used
IP_ADDRESS_DHCP -> MY_IP_ADDRESS_VIA_DHCP
```

I checked keywords.txt and none of the old defines have highlighting rules.

Avamander:
Yeah, probably would be a good idea. It probably wouldn't be a bad idea to just have a section in MyConfig.h that's for all the deprecated flags that throw errors.

mfalkvidd:
There is a list at https://www.mysensors.org/apidocs/deprecated.html but that page only lists stuff that have ben annotated as deprecated.

Avamander:
I more meant like that the file itself could have a separate section, so they aren't all mixed up with the rest, valid defined. I think that should be some other PR that does that though.

tekka007:
Closing after inactivity

PR #1429, opened 2020-06-06
Remove unnecessary MY_IP_RENEWAL_INTERVAL*

mfalkvidd:
Ethernet.maintain already checks when we need to renew, based on
the lease time given by the DHCP server.

https://www.arduino.cc/en/Reference/EthernetMaintain recommends
"just call it on every loop() invocation" so there is no need for
us to keep a separate timer.


core/MyGatewayTransportEthernet.cpp
```diff
@@ -150,19 +150,10 @@ void gatewayTransportRenewIP(void)
...
-
-	// http://playground.arduino.cc/Code/TimingRollover
-	if ((int32_t)(now - _nextIPRenewal) < 0) {
-		return;
-	}
 	if (Ethernet.maintain() & ~(0x06)) {
 		GATEWAY_DEBUG(PSTR("!GWT:TRC:IP RENEW FAIL\n"));
-		return;
 	}
 	_w5100_spi_en(false);
```
mfalkvidd: Not sure why `_w5100_spi_en(false)` was only called in case of successful renewal before. IF it should be called at all, my guess is that it should be called regardless whether renewal was successful or failed

Yveaux: gatewayTransportRenewIP is only called from gatewayTransportAvailable.
gatewayTransportAvailable starts with _w5100_spi_en(true);
With your change we return to gatewayTransportAvailable with SPE_EN set false.
This is likely to break the remaining code in gatewayTransportAvailable.

I would remove any call to _w5100_spi_en() from gatewayTransportRenewIP, so it never messes with the state of the SPI_EN line.

The SPI_EN handling in that file is horrible; I would almost advise to use gotos 😱

mfalkvidd: To be clear: without my change, we return to gatewayTransportAvailable with SPI_EN set to false when the renewal was successful (which it is most of the time).

mfalkvidd:
Ready for merge unless someone has any additional feedback.

mfalkvidd:
If we fail to renew repeatedly, we will spam the debug console with messages. Not sure if that will cause any problems.

PR #1436, opened 2020-08-18
added LGT8F support

arcayi:
Added architecture LGT8F.
LGT8F: https://github.com/dbuezas/lgt8fx
Remained issues:
LGT8F: can only run under 16Mhz, for the bootloader set it so.


tekka007:
Closing for now: no demand from community

PR #1445, opened 2020-10-18
NRF5: Fix ESB deadlock

### LINKED ISSUE #1371: NRF51822 hangs with heavy traffic and smartsleep
I've noticed using smartsleep with nrf51 platform, when heavy traffic is applied to the system (i.e. a lot of messages during switching groups), nrf51 hangs in Timer interrupt 'NRF5_RADIO_TIMER_IRQ_HANDLER' of Radio_ESB.cpp, with never cleared status of 'NRF5_RADIO_TIMER->EVENTS_COMPARE[1]'. This results in interrupt being handled over and over again i blocks execution. What I found in the code, this status should be handled using PPI channel. Clearing this status in debug resolves the issue until next such case occurs.
I found a solution in 'static bool NRF5_ESB_initialize()' changing radio and timer intterrupt priorities (basically swapping them) .With this, problem never occurs anymore.
Could someone take a look at this and verify the solution?

### LINKED ISSUE #1410: Deadlock in NRF5_ESB_sendMessage
Hi,

I see from time to time a deadlock in `NRF5_ESB_sendMessage()`:
https://github.com/mysensors/MySensors/blob/8103c6cf6bcb3e271df08b17d0fb2fee3b444c86/hal/transport/NRF5_ESB/driver/Radio_ESB.cpp#L470

For whatever reason, my nRF52840 doesn't get the interrupt, `events_end_tx` remains `false` and the `while` loop never terminates.

I'm using a nRF52840 Dongle, PlatformIO and a J-Link emulator.
I can debug the code, but I don't know Nordic chips, so any help is welcome.

d00616:
Fix for https://github.com/mysensors/MySensors/issues/1410

Under some conditions the NRF5 don't end the sending routine. This fix clears the timer used for timout and resending.

### DIFF
--- a/hal/transport/NRF5_ESB/driver/Radio_ESB.cpp
+++ b/hal/transport/NRF5_ESB/driver/Radio_ESB.cpp
@@ -179,17 +179,6 @@ static bool NRF5_ESB_initialize()
 	// Stop timer when CC0 reached
 	NRF5_RADIO_TIMER->SHORTS =
 	    TIMER_SHORTS_COMPARE3_CLEAR_Msk | TIMER_SHORTS_COMPARE3_STOP_Msk;
-	// Reset timer
-	NRF5_RADIO_TIMER->TASKS_CLEAR = 1;
-
-	// Reset compare events
-#ifdef NRF51
-	for (uint8_t i=0; i<4; i++) {
-#else
-	for (uint8_t i=0; i<6; i++) {
-#endif
-		NRF5_RADIO_TIMER->EVENTS_COMPARE[i] = 0;
-	}
 
 	// Enable interrupt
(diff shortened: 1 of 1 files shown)


mfalkvidd:
@d00616 is this ready for merge?

d00616:
@mfalkvidd Yes

mfalkvidd:
Great, thanks!

d00616:
I must create a new PR. After you have merged this PR my Gateways has stopped to send after 20 days of continuous operation. It looks like there was/is another error condition,

mfalkvidd:
Thanks. I have reverted the change.

d00616:
Reverting was a little bit to much. Both changes are required. I add both changes to https://github.com/mysensors/MySensors/pull/1448

PR #1447, opened 2020-11-07
WIP: NRF5: Fix ESB deadlock (rx)

d00616:
Part two of https://github.com/mysensors/MySensors/pull/1445, which doesn't fix all error conditions. I think there was a very little chance to run into an timer event while RX is active. Now timer interrupts are only active  in TX mode.

Next days, I review the code again.


d00616:
Replaced by https://github.com/mysensors/MySensors/pull/1455

Issue #1450, opened 2020-11-10
Auto id mode in RS485 transport doesn't work


klaudiusz223:
Automatic ID assignment doesn't work for RS485 transport.
As mentioned in this post https://forum.mysensors.org/topic/4854/building-a-wired-rs485-sensor-network/23 .
The problem is caused by:
- setting sender id to 0 instead of 255 in RS485 transport header when node doesn't have assigned ID
- wrong packet filtering in the RS485 transport header and discarding broadcast packets from nodes without assigned ID
https://github.com/mysensors/MySensors/blob/8f1f8ad2415034569a77d99f6ed9c02573c71418/hal/transport/RS485/MyTransportRS485.cpp#L149
https://github.com/mysensors/MySensors/blob/8f1f8ad2415034569a77d99f6ed9c02573c71418/hal/transport/RS485/MyTransportRS485.cpp#L168-L176

feanor-anglin:
Is this pull request (#1451) going to be merged any time soon?

PR #1451, opened 2020-11-10
RS485: Fix Auto ID assignment

klaudiusz223:
- fix for #1450 
- some code cleanup for RS485 (assertDE(), deassertDE() macros are now used)

### DIFF
--- a/hal/transport/RS485/MyTransportRS485.cpp
+++ b/hal/transport/RS485/MyTransportRS485.cpp
@@ -62,7 +62,7 @@
 #define deassertDE() hwDigitalWrite(MY_RS485_DE_PIN, LOW)
 #else
 #define assertDE() hwDigitalWrite(MY_RS485_DE_PIN, LOW); delayMicroseconds(5)
-#define deassertDE() hwDigitalWrite(MY_RS485_DE_PIN, HIG)
+#define deassertDE() hwDigitalWrite(MY_RS485_DE_PIN, HIGH)
 #endif
 #else
 #define assertDE()
@@ -260,14 +260,7 @@ bool transportSend(const uint8_t to, const void* data, const uint8_t len, const 
 		}
 	}
 
-#if defined(MY_RS485_DE_PIN)
-#if !defined(MY_RS485_DE_INVERSE)
-	hwDigitalWrite(MY_RS485_DE_PIN, HIGH);
-#else
-	hwDigitalWrite(MY_RS485_DE_PIN, LOW);
-#endif
-	delayMicroseconds(5);
-#endif
+	assertDE();
 
 	// Start of header by writing multiple SOH
 	for(byte w=0; w<MY_RS485_SOH_COUNT; w++) {
@@ -311,11 +304,7 @@ bool transportSend(const uint8_t to, const void* data, const uint8_t len, const 
 	_dev.flush();
 #endif
 #endif
-#if !defined(MY_RS485_DE_INVERSE)
-	hwDigitalWrite(MY_RS485_DE_PIN, LOW);
-#else
-	hwDigitalWrite(MY_RS485_DE_PIN, HIGH);
-#endif
+	deassertDE();
 #endif
 	return true;
 }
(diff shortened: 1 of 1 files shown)


hal/transport/RS485/MyTransportRS485.cpp
```diff
@@ -267,14 +260,7 @@ bool transportSend(const uint8_t to, const void* data, const uint8_t len, const
 		}
 	}
 
-#if defined(MY_RS485_DE_PIN)
```
mfalkvidd, line 270: Can we just stop checking MY_RS485_DE_INVERSE and MY_RS485_DE_PIN? Or do those checks be part of (de)assertDE?

klaudiusz223: This is crucial for correct transmission. If MY_RS485_DE_PIN is defined, it must be set to HIGH when transmitting and to receive anything it must be set to LOW. If we also define MY_RS485_DE_INVERSE , it means that this logic is inverted. If the pin is not defined, it means that switching between transmitting and receiving must be handled differently, e.g. by a transceiver. Most low-cost transceivers don't do that. 
BTW this part of pull request is just a code cleanup and has no effect on the object code or program operation. Just previously defined macros have been used.

mfalkvidd: This PR removes the two if clauses and replaces them with the macro. The macro does not contain the if clauses. Won't that result in non-working devices?

klaudiusz223: That's because this macro is defined differently depending on whether or not  MY_RS485_DE_PIN and MY_RS485_DE_INVERSE  are defined.
https://github.com/klaudiusz223/MySensors/blob/e70973144ee59f3c86616ff15f446fafa905340f/hal/transport/RS485/MyTransportRS485.cpp#L59-L70.

mfalkvidd: Great, thanks for explaining!

kasparsd:
Is there a way to get this merged? Without this fix the NODE ID discovery doesn't work at all when using the RS485 transport.

kasparsd:
Tried this using patch with an ESP8266 node and a Raspberry Pi controller (running the latest stable version 2.3.2) over RS485. Here is the output from the node attempting to register:

```
215 MCO:BGN:INIT NODE,CP=RSNNE---,FQ=80,REL=0,VER=2.4.0-alpha
445 TSM:INIT
459 TSF:WUR:MS=0
477 TSM:INIT:TSP OK
498 TSM:FPAR
511 TSM:FPAR:STATP=0
534 TSM:ID
545 TSM:ID:REQ
580 TSF:MSG:SEND,255-255-0-0,s=33,c=3,t=3,pt=0,l=0,sg=0,ft=0,st=OK:
2651 TSM:ID
2663 TSM:ID:REQ
2699 TSF:MSG:SEND,255-255-0-0,s=103,c=3,t=3,pt=0,l=0,sg=0,ft=0,st=OK:
4773 TSM:ID
4785 TSM:ID:REQ
4821 TSF:MSG:SEND,255-255-0-0,s=177,c=3,t=3,pt=0,l=0,sg=0,ft=0,st=OK:
6895 TSM:ID
6907 TSM:ID:REQ
6943 TSF:MSG:SEND,255-255-0-0,s=251,c=3,t=3,pt=0,l=0,sg=0,ft=0,st=OK:
9017 !TSM:ID:FAIL
9035 TSM:FAIL:CNT=1
9056 TSM:FAIL:DIS
9075 TSF:TDI:TSL
19094 TSM:FAIL:RE-INIT
19118 TSM:INIT
19133 TSM:INIT:TSP OK
19156 TSM:FPAR
19172 TSM:FPAR:STATP=0
19196 TSM:ID
19210 TSM:ID:REQ
```

and the matching messages received by the controller:

```
Jan 23 13:06:46 DEBUG TSF:MSG:READ,255-255-0,s=33,c=3,t=3,pt=0,l=0,sg=0:
Jan 23 13:06:48 DEBUG TSF:MSG:READ,255-255-0,s=103,c=3,t=3,pt=0,l=0,sg=0:
Jan 23 13:06:50 DEBUG TSF:MSG:READ,255-255-0,s=177,c=3,t=3,pt=0,l=0,sg=0:
Jan 23 13:06:52 DEBUG TSF:MSG:READ,255-255-0,s=251,c=3,t=3,pt=0,l=0,sg=0:
```

The node had the following extra flags:

```
#define MY_PARENT_NODE_ID 0 // Don't attempt to find it.
#define MY_PARENT_NODE_IS_STATIC // If the controller doesn't support returning the parent node.
```

~Is it supposed to be changing the sensor ID on every request?~ The presentation layer uses `#define CHILD_ID 1` and the following call `present(CHILD_ID, S_MOTION);`. Looks like the sensor ID is changing because of this:

https://github.com/mysensors/MySensors/blob/2e00bf6a10f76d6aaa1999e12313237bc3edabd3/core/MyTransport.cpp#L208-L214

which makes sense.

PR #1468, opened 2021-02-05
improve logparser for 2.x log format

zerog2k:
* prefilters log entries to strip out non-parseable data. This should handle arbitrary log line prefix data (e.g. full or millisecond timestamps, log levels, whitespacing, etc)
* removes old(?) log format assumptions
* output table rows monospace font for improve visual clarity

### DIFF
--- a/Logparser/logparser.html
+++ b/Logparser/logparser.html
@@ -5,8 +5,8 @@
 		<link rel="stylesheet" href="logparser.css" />
 	</head>
 	<body>
-		<script src="https://cdnjs.cloudflare.com/ajax/libs/vue/2.6.10/vue.min.js"></script>
-		<script src="https://cdnjs.cloudflare.com/ajax/libs/underscore.js/1.9.1/underscore-min.js"></script>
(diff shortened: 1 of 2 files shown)


zerog2k:
Ok I guess that the "old log format" assumptions is actually the core data output from serial gateway, which (probably?) should stay in? If so I can revise it. (Probably logparser needs to be revamped to display all the columns properly from parsing the debug output instead? It currently does not display any of the other fields properly when parsing typical debug output from nodes or rpi gateway pipe/log)

zerog2k:
ok yeah, i guess the ambiguity here is that it seems like the logparser attempts to handle two separate formats, and furthermore handling them differently, in different places of the Vue.js framework:
1. "serial protocol" output: handled by `parse` method - ref: https://www.mysensors.org/download/serial_api_20
2. debug "log" output: handled in `parse` when calling `match` (`decode`) method - ref: https://www.mysensors.org/apidocs/group__MyTransportgrp.html#details

So I guess continue supporting both formats for now?

zerog2k:
resubmitted to maintain serial protocol compatibilty in #1469

PR #1469, opened 2021-02-05
Improve logparser compatibility

zerog2k:
rework of #1468 to maintain compatibility with serial protocol parsing

* output table rows in monospace font to improve visual clarity

for debug log entries:
* prefilters to strip out non-parseable data. This should handle arbitrary log line prefix data (e.g. full or millisecond timestamps, log levels, whitespacing, etc)

IMHO, the table formatting is still not ideal as it's mixing display assumptions about serial protocol vs log entries. As discussed in #1468, will likely need overhaul to handle and display either type in a more generic way, but I'll leave that for another day.

### DIFF
--- a/Logparser/logparser.html
+++ b/Logparser/logparser.html
@@ -5,8 +5,8 @@
 		<link rel="stylesheet" href="logparser.css" />
 	</head>
 	<body>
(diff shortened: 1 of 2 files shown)


zerog2k:
here is an example dataset of entries I threw at it for testing:
https://gist.github.com/zerog2k/c403fc4c46361e7283cdacbe25b32fe5

PR #1470, opened 2021-02-10
Add internal battery measurement example

mfalkvidd:
For an update to https://www.mysensors.org/build/battery#measuring-and-reporting-battery-level based on work by skywatch in https://forum.mysensors.org/topic/11555/measuring-battery-voltage-which-way-is-best/

### DIFF
--- a/examples/BatteryPoweredSensorInternal/BatteryPoweredSensorInternal.ino
+++ b/examples/BatteryPoweredSensorInternal/BatteryPoweredSensorInternal.ino
@@ -55,7 +55,7 @@ void presentation()
 void loop()
 {
   // get the battery Voltage
-  long batteryMillivolts = readVcc();
+  long batteryMillivolts = hwCPUVoltage();
   int batteryPcnt = batteryMillivolts / FULL_BATTERY / 1000.0 * 100 + 0.5;
 #ifdef MY_DEBUG
   Serial.print("Battery Voltage: ");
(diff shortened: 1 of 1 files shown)


Yveaux:
I would suggest to use an external library (e.g. https://github.com/Yveaux/Arduino_Vcc) instaed.

mfalkvidd:
Do we have a good strategy for handling external dependencies?
Move this example to the MySensorsArduinoExamples repo, and copy the current version of your library into MySensorsArduinoExamples? That’s what we’ve done so far, but it isn’t a very good strategy imo.

Yveaux:
External libraries in Arduino are a real pain. PlatformIO handles it a lot better, allowing automatic installation of libraries, direct referencing and pinning specific versions. This is however a separate discussion...
In line with other examples this would require a copy of the Vcc library indeed.
Later we can decide how to improve on this in general.

mfalkvidd:
How about using `hwCPUVoltage()`? It seems to do something similar.

user2684:
For what it matters, on the NodeManager side I'm also using hwCPUVoltage()  (https://github.com/mysensors/NodeManager/blob/master/sensors/SensorBattery.h)

mfalkvidd:
Any additional comments before I merge?

Yveaux:
I suppose you tested it? If so, go ahead 👍🏻

examples/BatteryPoweredSensorInternal/BatteryPoweredSensorInternal.ino
```diff
@@ -0,0 +1,74 @@
...
+{
+  // Send the sketch version information to the gateway and Controller
+  sendSketchInfo("Battery Meter", "1.0");
+}
+
+void loop()
+{
+  // get the battery Voltage
+  long batteryMillivolts = hwCPUVoltage();
+  int batteryPcnt = batteryMillivolts / FULL_BATTERY / 1000.0 * 100 + 0.5;
```
tekka007, line 59: Why adding 0.5?

Yveaux: Round to next full int value

tekka007: i.e. full battery = 101%

Yveaux: Only if the calculation adds up to a value of 100.5 or more..
No need to blame the +0.5 😉

mfalkvidd:
> I suppose you tested it? If so, go ahead 👍🏻

I've tested with a 5V Pro Mini as serial gateway. It works fine:

```
16:25:59.014 -> Battery voltage: 4.48V
16:25:59.014 -> Battery percent: 149 %
```

I also tested with a nodemcu. It returns 0V, as expected since `hwCPUVoltage()` is not supported.

Yveaux:
```
16:25:59.014 -> Battery voltage: 4.48V
16:25:59.014 -> Battery percent: 149 %
```
That 149% is a bit funky, but that's likely because you didn't tune it or ran the pro mini from 3V, I suppose?

mfalkvidd:
> That 149% is a bit funky, but that's likely because you didn't tune it or ran the pro mini from 3V, I suppose?

I ran the example as-is, with `#define FULL_BATTERY 3`

Yveaux:
Then it makes total sense! 👍🏻

mfalkvidd:
On release: Change the battery page to point to the master branch instead of the development branch.

Issue #1439, opened 2020-08-19
N2N 


smartboxchannel:
How do I disable n2n? in some situations with exchange between nodes with echo, this causes some conflicts, mcu freezes.

tekka007:
There is currently no option to disable N2N traffic, however, you can comment this section to disable it:

https://github.com/mysensors/MySensors/blob/development/core/MyTransport.cpp#L548-L555

lelouch94:
I had the same issue and have followed the suggestion from @tekka007 and added a define for that. This is available in my branch.
[mwa_dev_Issue_1439](https://github.com/lelouch94/MySensors/tree/mwa_dev_Issue_1439)

But currently I am stuck with the CLA issue.

PR #1475, opened 2021-02-21
Adds option to disable N2N (#1439)

lelouch94:
This pull request adds the configuration option MY_TRANSPORT_N2N_FEATURE_DISABLED,
to disable the node to node transport feature.

see: [#1439](https://github.com/mysensors/MySensors/issues/1439)


core/MyTransport.cpp
```diff
@@ -553,6 +554,7 @@ bool transportRouteMessage(MyMessage &message)
 			}
 			TRANSPORT_DEBUG(PSTR("!TSF:RTE:N2N FAIL\n"));
 		}
+#endif
```
mfalkvidd, line 559: How about adding an else clause here, with something like
`TRANSPORT_DEBUG(PSTR("!TSF:RTE:N2N DIS\n"));` so people who enable this feature and forget about it (or don’t fully realize what it does) can get some information on why their N2N messages disappear?

MyConfig.h
```diff
@@ -1111,6 +1111,12 @@
 #define MY_TRANSPORT_DISCOVERY_INTERVAL_MS (20*60*1000ul)
 #endif
 
+/**
```
mfalkvidd, line 1114: Could you add a small note on when this feature can/should be used? Maybe some sort of use case description?

lelouch94: Hi @mfalkvidd, my use case is a node that is located in a dead zone and can only be reached via a repeater.
When the node sends telegrams to other nodes or the gateway, I always get the error message !TSF:RTE:N2N FAIL in the log. This is a waste of node battery energy and in addition it pollutes the airwaves with traffic that interferes with the communication of other nodes.

core/MyTransport.cpp
lelouch94: Okay, I will try it out and check how it works.

MyConfig.h
mfalkvidd: Do you know what causes the node to try to sent to other nodes?

lelouch94: Hi @mfalkvidd, I am intentionally sending the message to another actuator node which unfortunately is not directly reachable via N2N.

lelouch94:
Hi @mfalkvidd, I have implemented your suggested change and it works for me so far. I have omitted the '!' because it is not an error, but a setting. 

Are there any news regarding the CLA?

lelouch94:
Hi @mfalkvidd, I have now also extended the log parser for the new _N2N disabled_ message.

tekka007:
Closing after inactivity

PR #1480, opened 2021-03-13
Fix: Set CS pin on Linux (RaspberryPI)

pldemone:
With the commit 1c33609556e10b15079d8d7452cd1355457bb147 a change was introduced whereby CS is no longer set on Linux. Thus setting the option "--my-rf24-cs-pin=" has no function under Linux.

### DIFF
--- a/configure
+++ b/configure
@@ -197,25 +197,25 @@ function detect_machine {
             soc="BCM2837"
         fi
         ;;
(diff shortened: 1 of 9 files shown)


tekka007:
This PR targets the master branch which is not allowed. Please resubmit targetting the development branch.

PR #1499, opened 2021-08-18
ROCK Pi 4 (RK3399) support

docbender:


configure
```diff
@@ -248,7 +248,10 @@ function detect_machine {
         soc="AM33XX"
         ;;
     *)
-        soc="unknown"
```
mfalkvidd, line 251: Shouldn't we keep this (but in an else statement after trying to detect rockpi), in case the soc cannot be detected?

docbender: It can be kept there. However _soc_ variable is already initialized:
https://github.com/mysensors/MySensors/blob/eed25da65c6b8ba4c52262312e3b5304182ef859/configure#L154-L155

PR #1501, opened 2021-08-24
Bridge transport implementation

lukfi:
I have implemented a new type of gateway transport for devices (Yún) that use Bridge library (https://www.arduino.cc/en/Reference/YunBridgeLibrary) for ethernet communication. I added a new define - MY_GATEWAY_BRIDGE - which enables the bridge communication. I have tested the changes in my local environment (I use seeeduino cloud) and didn't find any issues. The bridge transport works in server mode, in future it can be extended to support client mode.


tekka007:
Closing after inactivity, please feel free to re-open if needed

PR #1510, opened 2021-10-29
Add SoftwareSerial support for RS485 transport

### LINKED ISSUE #1300: MinorChanges: SoftwareSerial for RS485
Todo:
+) include SoftwareSerial to drivers
+) include espsoftwareserial to drivers

Changes in File MySensors.h

replace line 372 with
----
#if defined(MY_RS485_SWSERIAL)
#if defined(ESP8266) || defined(ESP32)
#include "drivers/ESPSoftwareSerial/SoftwareSerial.cpp"
#elif 
#include "drivers/SoftwareSerial/SoftwareSerial.cpp"
#endif 
#elif 
#include "drivers/AltSoftSerial/AltSoftSerial.cpp"
#endif 
-------

File hal/transport/RS485/MyTransportRS485.cpp
replace line 90 with:
-----
#if defined(MY_RS485_SWSERIAL)
SoftwareSerial& _dev  = MY_RS485_SWSERIAL;
#elif 
#include "drivers/AltSoftSerial/AltSoftSerial.cpp"
#endif 
------

thank you!

kasparsd:
Fixes #1300.

The Arduino [SoftwareSerial](https://www.arduino.cc/en/Reference/softwareSerial) library (and its forks for various hardware platforms) appears to be compatible with the API currently used by the [RS485 transport](https://github.com/mysensors/MySensors/blob/aa76d266159d87cd460597657ac7f669a706fad5/hal/transport/RS485/MyTransportRS485.cpp). It contains the following calls:

- `.available()`
- `.read()`
- `.flush()`
- `.begin()`


kasparsd:
Would this be something valuable for the upstream project? The `AltSoftSerial` implementation currently doesn't support ESP devices so this would solve that issue.

Issue #1527, opened 2022-07-05
PJON driver produces warning and lets fail CI and Toll Gate


virtual-maker:
There is a problems with the newly included PJON library.
The addition seems to break the CI and let always fail the "Toll gate" step.

Details:
Toll gate (STM32F1 - Tests) — Warnings found

From Jenkins log:
```
In file included from /var/lib/jenkins/jobs/MySensors/jobs/MySensors/branches/PR-1488/workspace/MySensors/hal/transport/PJON/driver/PJON.h:62,
                 from /var/lib/jenkins/jobs/MySensors/jobs/MySensors/branches/PR-1488/workspace/MySensors/MySensors.h:397,
                 from /var/lib/jenkins/jobs/MySensors/jobs/MySensors/branches/PR-1488/workspace/MySensors/tests/Arduino/sketches/pjon_transport/pjon_transport.ino:24:
/var/lib/jenkins/jobs/MySensors/jobs/MySensors/branches/PR-1488/workspace/MySensors/hal/transport/PJON/driver/PJONDefines.h: In static member function 'static void PJONTools::parse_header(const uint8_t*, PJON_Packet_Info&)':
/var/lib/jenkins/jobs/MySensors/jobs/MySensors/branches/PR-1488/workspace/MySensors/hal/transport/PJON/driver/PJONDefines.h:415:31: warning: 'void* memset(void*, int, size_t)' clearing an object of non-trivial type 'struct PJON_Packet_Info'; use assignment or value-initialization instead [-Wclass-memaccess]
  415 |   memset(&info, 0, sizeof info);
      |                               ^
/var/lib/jenkins/jobs/MySensors/jobs/MySensors/branches/PR-1488/workspace/MySensors/hal/transport/PJON/driver/PJONDefines.h:207:8: note: 'struct PJON_Packet_Info' declared here
  207 | struct PJON_Packet_Info {
      |        ^~~~~~~~~~~~~~~~
```
I assume that this `memset(&info, 0, sizeof info);` is not best practice and should be replaced.

PR #1488, opened 2021-05-03
Add CAN transport layer

AdamSlowik:
Add new transport layer.
Uses github.com/coryjfowler/MCP_CAN_lib maintained by @coryjfowler.

PR #1528, opened 2022-07-05
Replace memset() in PJONDefines.h to avoid build warning (#1527)

virtual-maker:
Fix for Issue #1527

This is a replacement for the `memset(&info, 0, sizeof info)` line, which causes warnings and fails the Toll Gate. Solution is to use `info = PJON_Packet_Info{};` instead.

This solution is taken from the latest PJON version at:
https://github.com/gioblu/PJON/blob/master/src/PJONDefines.h#L398
Latest commit [1be4b15](https://github.com/gioblu/PJON/commit/1be4b1587014180da3393e88c2458bbd0e8e9d22)

### DIFF
--- a/hal/transport/PJON/driver/PJONDefines.h
+++ b/hal/transport/PJON/driver/PJONDefines.h
@@ -412,8 +412,9 @@ struct PJONTools {
 
 	static void parse_header(const uint8_t *packet, PJON_Packet_Info &info)
 	{
-		memset(&info, 0, sizeof info);
 		uint8_t index = 0;
+		info = PJON_Packet_Info{};
+
 		info.rx.id = packet[index++];
 		bool extended_length = packet[index] & PJON_EXT_LEN_BIT;
 		info.header = packet[index++];


mfalkvidd:
Should (can?) we update the entire Pjon library from upstream (instead of this single change)?

virtual-maker:
> Should (can?) we update the entire Pjon library from upstream (instead of this single change)?

@mfalkvidd Hi Mikael,
Yes, upgrading the PJON Library is a good idea and should be done. And yes, the upgrade can potentially introduce new problems.

I personally don't use MySensors with PJON. So I can't judge if the new version really works with MySensors. So I think a PR with the upgrade to the latest PJON version should be requested by a PJON user who can test this with his use case.

@gryzli133 is it possible for you to create a PR with an update to the latest PJON version?

I personally find the CAN bus extension of @AdamSlowik PR #1488 very interesting and would be happy if it will be included in MySensors. 

Unfortunately Toll Gate does not accept the pull request as described. And also several other PRs, e.g. @Tico06 with PR #1520, currently have the same problem and are blocked.

Therefore, my suggestion is to first merge this PR which (hopefully) do not changes the current PJON functionality, so that the other PRs are no longer blocked. Afterwards, the PR of the PJON Library upgrade can be placed and tested at leisure.

BR Immo

Tico06:
Hi There,

Thanks a lot @virtual-maker for the fix and @mfalkvidd for the merge. I just commited let's see what will happen.

Thx and br

Eric.

PR #1537, opened 2022-12-30
make it possible to use external wifi manager 

magwas:
... by defining MY_EXTERNAL_WIFIMANAGER_USED

This makes possible to use the WifiManager library by not insisting on SSID to be defined, and not starting the wifi from the transport.


mfalkvidd:
Thanks for your contribution @magwas 
Could you please read the butler report at https://ci.mysensors.org/job/MySensors/job/MySensors/job/PR-1537/1/The_20Butler_20report/butler.html ? Let us know if you need any assistance.

magwas:
I believe that the cppcheck issue is not mine. If you want me to fix it, I can, though probably the code complexity does not worth that two bytes of RAM even in an embedded system.

magwas:
Okay, I fixed it nevertheless. Clean report is important, and disabling the warning would have been nearly as many code as fixing it the right way. And I had to look up how, so more work :)

PR #1545, opened 2023-03-16
Support for ASR650x architecture and SX126x transport

eiten:
I added support for the ASR650x architecture which can be found in the cheap Heltec Cubecell or the Ebyte E78-470LN22S modules.
As the ASR650x has an SX126x compatible transport, I added this transport. It also works for stand alone SX126x chips, for example the Heltec HT-RA62 together with an ESP32.


eiten:
Can anyone help me with this buttler message? Thanks!

PR #1547, opened 2023-04-14
Added support for ASR650x and SX126x

eiten:
As the tile states, added support for ASR6501, ASR6502 and the SX126x transport. Fixed up git now.


tekka007:
Changes & Fixes implemented in #1548

PR #1562, opened 2024-03-13
Bug fix for GPIO under recent Linux kernel versions

romaricr:
Slight update of MySensors using libgpiod instead of all standard sysfs to map interrupts.


romaricr:
Tested on my Raspberry PI / NRF24 gateway and working fine (with around 10 devices). Not tested on any other platform.

mfalkvidd:
Nice work, thanks!

Attaching the restyling that the butler requests: 
[restyling.patch](https://github.com/mysensors/MySensors/files/14623973/restyling.patch)

Will add further comments shortly.

MyConfig.h
```diff
@@ -442,7 +442,7 @@
  * - RF24_PA_MAX = 0dBm
  */
 #ifndef MY_RF24_PA_LEVEL
-#define MY_RF24_PA_LEVEL (RF24_PA_HIGH)
```
mfalkvidd: I'd prefer if we didn't change the default PA level in this PR, since it is not related to the change.

hal/architecture/Linux/drivers/core/interrupt.cpp
```diff
@@ -239,13 +221,13 @@ void detachInterrupt(uint8_t gpioPin)
...
-	}
-	fprintf(fp, "%d", gpioPin);
-	fclose(fp);
+//	FILE *fp = fopen("/sys/class/gpio/unexport", "w");
+//	if (fp == NULL) {
+//		logError("Unable to unexport pin %d for interrupt\n", gpioPin);
+//		exit(1);
+//	}
+//	fprintf(fp, "%d", gpioPin);
+//	fclose(fp);
```
mfalkvidd: Would you mind deleting those lines instead of commenting them? We have git for history.

hal/architecture/Linux/drivers/core/interrupt.h
```diff
@@ -24,6 +24,9 @@
 
 #include <stdint.h>
 
+// Ajout RRO
+#include <gpiod.h>
```
mfalkvidd, line 27: Do you know if this is installed by default on Raspberry Pi OS? If it is not, we should probably document how to install the dependency.

mfalkvidd (commented):
Adding review comments.

Do you know how long this interface has been available? The reason I'm asking is that we would prefer to not break older gateways (running on older Linux versions) if people upgrade.

romaricr:
Styles and minor updates done.
For libgpiod, it has been there for a while. The [release notes says 6 years](https://git.kernel.org/pub/scm/libs/libgpiod/libgpiod.git).
About the potential dependency, you are right. It might require a `sudo apt-get install libgpiod-dev`  to build and `sudo apt-get install libgpiod` to run. I have not tested what it looks like on a genuine installation.

For the documentation, sorry for my question but, how do I do that ?

mfalkvidd:
Excellent, thanks!

6 years should be good enough, imo.

The documentation is not under version control unfortunately but I can add it to https://www.mysensors.org/build/raspberry

romaricr:
Thank you to the reviewers. 
I have seen some of the errors but I'm really not sure how to fix them ... Anyone could help ?

pompilir:
Hi, 
I was trying the fixed version after upgrading my RPI to the latest Raspbian OS. However, even after using the MySensors-gpio_fix version of the library I get "Could not open /sys/class/gpio/gpio16/direction"
Can you please help me?
Thanks in advance!

functionpointer:
Just tested this on RPi Zero W with Raspbian Bookworm 32-bit kernel. Works great.

raks-dev:
> Hi, I was trying the fixed version after upgrading my RPI to the latest Raspbian OS. However, even after using the MySensors-gpio_fix version of the library I get "Could not open /sys/class/gpio/gpio16/direction" Can you please help me? Thanks in advance!

What is the SOC and CPU detected as part of the configure command execution?

E.g. 
```
[SECTION] Detecting target machine.
  [OK] machine detected: SoC=BCM2712, Type=rpi5, CPU=aarch64.
[SECTION] Detecting SPI driver.
  [OK] SPI driver detected:BCM.
```

update:
I have proposed the solution which worked for me here: https://github.com/mysensors/MySensors/issues/1567

cvdenzen:
like @raks-dev and @ricgyver I also still see the error "Could not open /sys/class/gpio/gpio9/direction". It seems that this pull request is not complete. The error is coming from file "hal/architecture/Linux/drivers/core/GPIO.cpp".
Anyone who can fix it (or maybe someone already has fixed it)?

raks-dev:
@cvdenzen Refer this https://github.com/mysensors/MySensors/issues/1567
I have proposed the solution which worked for me.

cvdenzen:
> @cvdenzen Refer this #1567 I have proposed the solution which worked for me.
@raks-dev @ricgyver 
The proposed solution doesn't work for me. I think the difference is that I am using a rpi4 and RFM69, and not the NRF24. The RFM69 code tries to set the interrupt pin to input mode. And that will result in the error "Could not open /sys/class/gpio/gpio16/direction".
I have a working solution, see https://github.com/cvdenzen/MySensors/tree/carl-gpiod1.

This solution includes:
- detection of a raspberry Pi 4 under Linux 6.6 (in the configure script)
- the edits by @romaricr .

Edit: I don't think the GPIO files are used with the changed configuration file (to detect the rpi4). I don't understand it, but at least it works for me.

hal/architecture/Linux/drivers/core/interrupt.cpp
```diff
@@ -153,72 +154,53 @@ void attachInterrupt(uint8_t gpioPin, void (*func)(), uint8_t mode)
 		usleep(1000);
 	}
```
poehlert, line 156: 143-146 are now all unused vars from deleted code (fd, fName, c, count)

hal/architecture/Linux/drivers/core/interrupt.cpp
```diff
@@ -153,72 +154,53 @@ void attachInterrupt(uint8_t gpioPin, void (*func)(), uint8_t mode)
...
-	// Export pin for interrupt
-	if ((fd = fopen("/sys/class/gpio/export", "w")) == NULL) {
-		logError("attachInterrupt: Unable to export pin %d for interrupt: %s\n", gpioPin, strerror(errno));
-		exit(1);
-	}
-	fprintf(fd, "%d\n", gpioPin);
-	fclose(fd);
+	char *chipname = "gpiochip0";
+	unsigned int line_num = gpioPin;
+	struct gpiod_line_event event;
```
poehlert, line 159: this is unused as well

pompilir:
After some time I was finally able to get MySensors running on a RPI3B serial gateway with bookworm os. I had to use this gpio_fix version of the library (the current dev branch does not compile). It is working reliably by almost a week with the sensors network I build in previous years. As steted, I had to install the libgpiod-dev library and use the following config lines:

./configure --my-transport=rfm69 --my-rfm69-frequency=433 --my-is-rfm69hw --my-gateway=serial --my-serial-is-pty --my-serial-pty=/dev/ttyUSB020 --soc=BCM2837 --cpu-flags="-march=armv8-a+crc -mtune=cortex-a53"_

All the other attempts (without the --soc=BCM2837 --cpu-flags="-march=armv8-a+crc -mtune=cortex-a53 instructions)  failed. Hope this can push the merge of the fix in the official release of MySensors.

zkarolyi:
For make this needed: `sudo apt-get install libgpiod-dev`

hal/architecture/Linux/drivers/core/interrupt.h
mfalkvidd: `sudo apt-get install libgpiod-dev`
Thanks to @zkarolyi  in https://github.com/mysensors/MySensors/pull/1562#issuecomment-3193569605

andreacioni:
Just to confirm that I tried base `development` merged with `romaric/gpio_fix` and the build completed flawless. However this two additions are mandatory if you want this to work:

- `sudo apt-get install libgpiod-dev`
- configure flags: `--soc=BCM2837 --cpu-flags="-march=armv8-a+crc -mtune=cortex-a53`

Considering the relevance of this for RPi 5 user it should be documented somewhere.

zkarolyi:
For RPi 4 I have these **WORKING**:

- base `development `merged with `romaric/gpio_fix`
- `sudo apt-get install libgpiod-dev`
- `sudo ./configure --my-gateway=mqtt --my-controller-ip-address=127.0.0.1 --my-mqtt-publish-topic-prefix=mysensors-out --my-mqtt-subscribe-topic-prefix=mysensors-in --my-mqtt-client-id=mygateway1 --my-mqtt-user=<...> --my-mqtt-password=<...> --soc=BCM2711 --cpu-flags="-march=armv8-a+crc -mtune=cortex-a72"`

ONLY for **IRQ**:
But after a little fix in code to be able to `make` it with IRQ:
In `/hal/architecture/Linux/MyHwLinuxGeneric.h`:
from:
```
static __inline__ void __hwLock()
{
	pthread_mutex_lock(&hw_mutex);
}
```
to:
```
static __inline__ uint8_t __hwLock()
{
	pthread_mutex_lock(&hw_mutex);
	return 1;
}
```
I can not use IQR (`--my-rf24-irq-pin=22`). So working without it then.

PR #1486, opened 2021-04-17
Arduino_core_STM32 support

Elektroman2:
Is there any chance to merge Arduino_core_STM32 support to the main branch?

https://forum.mysensors.org/topic/10998/stm32l0-support/5?_=1618699197788

PR #1437, opened 2020-08-18
Stm32duino

arcayi:
Added architecture STM32duino. The STM32duino support more stm32 boards than Roger's core which only support STM32F1 so far to my knowledge.
STM32duino: https://github.com/stm32duino/Arduino_Core_STM32
Remained issues:
STM32duino: sleep mode not implemented yet.

PR #1422, opened 2020-05-24
Add STM32 cores support

KooLru:
This pull request adds initial support for STM32 Cores https://github.com/stm32duino/Arduino_Core_STM32 officially developed Arduino STM32 boards package from STMicroelectronics. It is based on HAL and LL libraries. 

It has been tested with the Blue Pill (STM32F103C8T6).

PR #1571, opened 2025-10-21
Arduino_core_STM32 support (improved)

dirkju:
Adding Arduino_core_STM32 support

Three previous attempts of adding STM32 support were abandoned (#1422, #1437, #1486).
@tekka007 s [branch ](https://github.com/tekka007/MySensors/commit/ab0c5e25163a22d3831f92df41b993dc70c5075c#diff-4e6427b267666a8140d595b1ca5bf0fd94aad38b9ce8a988f0e57d14db5c22bf) also served as inspiration. Thanks to everyone of you.

I'm asking you to please consider my attempt :-) @tekka007 @mfalkvidd 

### Risks
The change is rather local in hal/STM32 and should not cause problems with other architectures.
I've kept the existing legacy hal/STM32F1 unchanged and the respective constant is checked in MySensors.h before my STM32 addition. This should maximize backward-compatibility for STM32F1 (blue pill) legacy users.

### Implemented in this version
- [x] Serial communication (USB CDC and Hardware UART)
- [x] SPI interface for radios
- [x] EEPROM emulation using Flash memory
- [x] Watchdog support (requires explicit initialization)
- [x] System reboot
- [x] Random number generation (using internal temperature sensor)
- [x] Unique device ID (96-bit STM32 UID)
- [x] CPU voltage reading (via VREFINT)
- [x] CPU temperature reading (via internal sensor)
- [x] CPU frequency reporting
- [x] Critical section (interrupt disable/restore)
- [x] RAM routing table support

### Planned
- [ ] Low-power sleep modes (STOP, STANDBY)
- [ ] RTC-based timekeeping
- [ ] Interrupt-based wake from sleep
- [ ] Free memory reporting (heap analysis)

### Build
I developed under VS Code + Platformio on Windows. I test-compiled examples under the Arduino IDE (also Windows).

### Testing
I've tested mostly on a generic black pill STM32F401CE (gateway, node using Ra-01 LoRa and nRF24L01+).
I encourage to test on a broader variety of boards.

### Final note 
The code has mostly been generated using AI. I asked for a comprehensive review of the previously mentioned three abandoned STM32 PRs and a comparison between building on STM32duino vs the original STM32Cube HAL. Both are feasible but the simplicity of the STM32duino solution stood out.


tekka007:
@dirkju Thank you for this PR. Please see my review comments that need some attention.

hal/architecture/STM32/MyHwSTM32.cpp
```diff
@@ -0,0 +1,308 @@
...
+#ifdef IWDG
+	// Reset independent watchdog if enabled
+	// Note: Watchdog must be configured separately in sketch if needed
+	#if defined(HAL_IWDG_MODULE_ENABLED)
+	// Using STM32 HAL
+	// Implementation depends on whether user has initialized IWDG
+	// For safety, we only reset if it's running
+	#endif
+#endif
+	// No-op if watchdog not enabled - safer default
```
tekka007: Here some code is missing to actually reset the wdt if enabled

hal/architecture/STM32/MyHwSTM32.cpp
```diff
@@ -0,0 +1,308 @@
...
+#endif
+}
+
+uint16_t hwCPUVoltage(void)
+{
+#if defined(AVREF) && defined(__HAL_RCC_ADC1_CLK_ENABLE)
+	// Read internal voltage reference to calculate VDD
+	// VREFINT is typically 1.2V (varies by STM32 family)
+
+	uint32_t vrefint = analogRead(AVREF);
```
tekka007, line 192: AVREF is accessed without checking the VREF_AVAILABLE flag, see above in hwRandomNumberInit()

hal/architecture/STM32/MyHwSTM32.cpp
```diff
@@ -0,0 +1,308 @@
...
+	return F_CPU / 100000UL;
+}
+
+int8_t hwCPUTemperature(void)
+{
+#if defined(ATEMP) && defined(__HAL_RCC_ADC1_CLK_ENABLE)
+	// Read internal temperature sensor
+	// Note: Requires calibration values for accurate results
+
+	int32_t temp_raw = analogRead(ATEMP);
```
tekka007, line 219: ATEMP is accessed without checking the TEMP_SENSOR_AVAILABLE flag

hal/architecture/STM32/MyMainSTM32.cpp
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * @file MyMainSTM32.cpp
+ * @brief Main entry point for STM32
+ *
+ * STM32duino core provides main() function and serialEvent() handlers.
+ * No additional implementation needed - the framework handles setup()/loop() calls.
+ */
+
+// Note: STM32duino core already provides weak serialEvent handlers
+// We don't need to redefine them here
```
tekka007: Hook to _begin() and _process() is missing, see other HAL implementation for reference

tekka007 (changes requested):
Thanks @dirkju, please see my comments. I believe it makes sense to merge the (old/legacy) STM32F1 implementation to have one unified STM32 - if possible. What do you think?

hal/architecture/STM32/MyHwSTM32.cpp
dirkju: added.

hal/architecture/STM32/MyHwSTM32.cpp
dirkju: added condition

hal/architecture/STM32/MyHwSTM32.cpp
dirkju: added condition

hal/architecture/STM32/MyMainSTM32.cpp
dirkju: added.

dirkju:
> Thanks @dirkju, please see my comments. I believe it makes sense to merge the (old/legacy) STM32F1 implementation to have one unified STM32 - if possible. What do you think?

@tekka007 Completely agree, this is my goal. I was not sure though, if we should do that in a single step. Wdyt - should we try to get out the 2.4.0 release with both STM32 and /F1 and thereafter produce a 2.5 with STM32F1 removed?

PR #1586, opened 2025-12-23
STM32 sleep support

dirkju:
Support for MySensors sleep mode for STM32 using STOP Mode.

PR #1592, opened 2026-02-14
Add STM32U0 RTC support for sleep (extends #1586)

nschurando:
The STM32U0 series uses ICSR instead of ISR for the RTC initialization status register, and has a combined RTC_TAMP_IRQn instead of a separate RTC_WKUP_IRQn. Add a dedicated init block so sleep/wakeup works correctly on this family.


tekka007:
Thanks, @nschurando, could you please rebase your commit onto the current dev branch?

nschurando:
@tekka007, done :)

PR #1599, opened 2026-03-21
Add STM32WL Support to hal architecture

spike314:
Some Arduino Core STM32WL HAL callouts differ enough from the STM32 processors currently supported by MySensors that it causes compile and function errors.
This PR changes the MySensors STM32 architecture hal to address the STM32WL differences. 

While testing these updates, I identified a latent issue with STM32F1 sleep.  After waking the first time, then going back to sleep, the STM32F1 would not wake from timer again.  This PR also addresses the sleep issue.

The newer STM32 processors have millisecond resolution real time clocks.  This PR adds support for those clocks in the hwGetSleepRemaining function.  Newer processors will return sleep remaining within a few milliseconds.   The STM32F1 only has 1 second resolution.  It's behavior is unchanged, reporting in milliseconds, but only to the nearest second.

Addresses issues #1600 and #1601

I tested these changes on:
     STM32WLE5  (WIO-E5 and WIO-E5-LE modules)
     STM32F411CE (black pill)
     STM32F103 (blue pill)

### DIFF
--- a/hal/architecture/STM32/MyHwSTM32.cpp
+++ b/hal/architecture/STM32/MyHwSTM32.cpp
@@ -477,11 +477,6 @@ static bool hwSleepInit(void)
 	EXTI->PR = EXTI_PR_PR17;
 	NVIC_ClearPendingIRQ(RTC_Alarm_IRQn);
 
-	// Configure EXTI line 17 for RTC Alarm wake-up from STOP mode
-	// Without EXTI, the alarm flag is set but the CPU does not wake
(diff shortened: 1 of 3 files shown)


tekka007:
Thanks @spike314 👍

spike314:
Thanks @tekka007 .  I am working on an update to the SX126x driver to make it work with the STM32WL processors, since they have a built in radio.  The updated driver is "mostly" working.  I'll submit a PR for that one when I get the last bugs squashed.

PR #1603, opened 2026-04-10
enable broadcast messages in node

CodeGarageIo:
enable broadcast messages in node

### DIFF
--- a/core/MyTransport.cpp
+++ b/core/MyTransport.cpp
@@ -699,7 +699,8 @@ void transportProcessMessage(void)
 	_transportSM.msgReceived = true;
 
 	// Is message addressed to this node?
-	if (destination == _transportConfig.nodeId || (destination == BROADCAST_ADDRESS && command == C_SET)) {
+	if (destination == _transportConfig.nodeId || (destination == BROADCAST_ADDRESS &&
+	        command == C_SET)) {
 		// null terminate data
 		_msg.data[msgLength] = 0u;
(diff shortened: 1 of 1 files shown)


core/MyTransport.cpp
```diff
@@ -699,7 +699,8 @@ void transportProcessMessage(void)
 	_transportSM.msgReceived = true;
 
 	// Is message addressed to this node?
-	if (destination == _transportConfig.nodeId) {
+	if (destination == _transportConfig.nodeId || 
+		(destination == BROADCAST_ADDRESS && (command == C_SET || command == C_REQ))) {
```
tekka007: Broadcast messages are processed here, better to add the additional logic in that section: https://github.com/mysensors/MySensors/pull/1603/changes#diff-a02a87fdbafded90ea048bd15ab0c7d7d623c292098d2ab80b56faa7709f299aL853

CodeGarageIo: It will end with code repetition. And we will get two different entry points to receive() in one flow.

tekka007: Yeah for sure. To better understand, what (type of) messages are to be sent to all nodes?

CodeGarageIo: For example, you have multiple nodes with S_COVER and you send broadcast with V_DOWN to close it all at once, removing tons of messages from transport layer. In home automation it is very very common to use broadcast.

PR #1604, opened 2026-04-10
disable discovery if GW use RS485 transport

CodeGarageIo:
Disable MY_TRANSPORT_DISCOVERY if RS485 transport layer used.


tekka007 (changes requested):
This disables network discovery for all non-RS485 transports

CodeGarageIo:
Yes, sorry I missed negation during copy&paste operation ;)

tekka007 (approved):

PR #1608, opened 2026-04-11
Replace memcpy with memmove for buffer handling

tekka007:
Overlapping read/write in memcpy() is undefined behavior

### DIFF
--- a/hal/transport/RS485/MyTransportRS485.cpp
+++ b/hal/transport/RS485/MyTransportRS485.cpp
@@ -145,8 +145,8 @@ bool _serialProcess()
 		// our ID or BROADCAST_ADDRESS, save the header information and progress to
 		// the next state.
 		case 0:
-			for (uint8_t i = 0; i < 5; i++) {
-				_header[i] = _header[i+1];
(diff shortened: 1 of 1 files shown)


hal/transport/RS485/MyTransportRS485.cpp
```diff
@@ -145,7 +145,7 @@ bool _serialProcess()
 		// our ID or BROADCAST_ADDRESS, save the header information and progress to
 		// the next state.
 		case 0:
-			memcpy(&_header[0],&_header[1],5);
+			memmove(&_header[0],&_header[1],5);
```
CodeGarageIo: Consider leaving memcpy here, it is 1,5 to 2 times faster than memmove(). And in this particular case memcpy() works in forward copy mode (dst address < src address), which is overlap safe. 
If you afraid of LTO or other compiler mambo-jambo, better option is to write this code explicitly. Somthing like:
for (uint8_t i = 0; i < 5; i++) {
    _header[i] = _header[i+1];
}
it will be as fast and light as memcpy, and apparently overlap safe.

PR #1609, opened 2026-04-11
Linux: Fix compilation with EthernetServer.cpp

functionpointer:
As of https://github.com/mysensors/MySensors/pull/1585 [MyGatewayTransportEthernet.cpp](https://github.com/mysensors/MySensors/pull/1585/changes#diff-eeff1bb5cd3c9bcf20ea918cfaa75d72c4258e06410be42aa848ff51ec322453) uses `accept()` instead of `available()`.

Therefore for Linux we should provide `accept()`.


tekka007:
Thanks!

PR #1610, opened 2026-04-12
Linux: Migrate to libgpiod 2.x

functionpointer:
Building MySensors on Raspbian Trixie does not work, because the sysfs interface uses strange pin numbers now.
Furthermore, Trixie ships with libgpiod 2.x, so https://github.com/mysensors/MySensors/pull/1562 does not work either (it expects libgpiod 1.x).

This PR proposes to move from sysfs interface directly to libgpio 2.x, enabling MySensors to work on Trixie and hopefully beyond.
To that end, the hopefully more stable core libgpiod api is used in favor of C++ bindings.

Building MySensors for Linux requires `sudo apt-get install libgpiod-dev`

Tested on Raspberry Pi Zero W with RFM69HCW.


functionpointer:
The CI suggested some additional style changes in unrelated files. Those should probably happen outside this PR.

PR #1612, opened 2026-04-13
Ethernet lib >= 2.0.0

CodeGarageIo:
Ethernet lib since version 2.0.0 handle W5100/W5500 SS pin by itself. So MY_W5100_SPI_EN is now mandatory.

### DIFF
--- a/core/MyGatewayTransportEthernet.cpp
+++ b/core/MyGatewayTransportEthernet.cpp
@@ -126,24 +126,9 @@ static EthernetClient clients[MY_GATEWAY_MAX_CLIENTS];
 static inputBuffer inputString;
 #endif /* End of MY_GATEWAY_CLIENT_MODE */
 
-// On W5100 boards with SPI_EN exposed we can use the real SPI bus together with radio
-// (if we enable it during usage)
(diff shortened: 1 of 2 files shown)


core/MyGatewayTransportEthernet.cpp
```diff
@@ -126,24 +126,9 @@ static EthernetClient clients[MY_GATEWAY_MAX_CLIENTS];
...
-		hwPinMode(MY_W5100_SPI_EN, INPUT);
-		hwDigitalWrite(MY_W5100_SPI_EN, HIGH);
-	} else {
-		// Ground pin
-		hwPinMode(MY_W5100_SPI_EN, OUTPUT);
-		hwDigitalWrite(MY_W5100_SPI_EN, LOW);
-	}
-#else
-	(void)enable;
+#if !defined(MY_W5100_SPI_EN)
```
tekka007: Wrap this directive in a conditional check for MY_GATEWAY_W5100 to prevent build failures on other gateway types.

tekka007 (changes requested):

tekka007 (approved):

